Issue #590, opened 2014-04-23
Client performing introspection needs all scopes the token was issued with


alimfeld:
We have a setup with multiple resource servers and one relying party accessing those resource servers. The relying party gets an access token (issued to the relying party client) to do so. The access token has multiple custom scopes associated with it. We use a custom scope per resource server the relying party accesses. The resource server then uses the introspection endpoint to introspect the access token and maps the returned scopes to granted authorities. Every resource server allows access only in case the custom scope defined for this specific resource server is associated with the access token.

As token introspection is only allowed in case the client performing the introspection is the same client that the token was issued to, or it at least has all the scopes the token was issued with (see https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/3b52ce82019b1e2c917c91d4d01b153039ccf297/openid-connect-server/src/main/java/org/mitre/oauth2/web/IntrospectionEndpoint.java#L122), we are forced to associate all those custom scopes with every resource server client (every resource server performs the introspection as a specific resource server client).

What are your thoughts on this setup? Is it possible to override the behavior of the IntrospectionEndpoint (i.e. don't require the introspecting client to have all the scopes of the access token; and what are the security implications)? Should we use separate access tokens per resource server? Did we completely misunderstand/misuse the whole concept of scopes?

Any hints about this issue are much appreciated!

PR #598, opened 2014-05-07
Make introspection endpoint access authorization pluggable.

alimfeld:
This allows us to plug in our own access authorization logic to use the introspection endpoint (see https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/issues/590).

We can do so by having our own implementation of the IntrospectionAuthorizer interface and marking it as the primary bean for autowiring (i.e. using the @Primary annotation).

### DIFF
--- a/openid-connect-common/src/main/java/org/mitre/oauth2/service/IntrospectionAuthorizer.java
+++ b/openid-connect-common/src/main/java/org/mitre/oauth2/service/IntrospectionAuthorizer.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright 2014 The MITRE Corporation
+ *   and the MIT Kerberos and Internet Trust Consortium
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


jricher:
@alimfeld : Just a heads up: I'm needing to rewrite this for the UMA branch and I'm considering changing the whole way that introspection scopes are handled in the process. Current thinking is that a client can introspect any token but only get back the scopes that it's registered for. In the UMA case (it authenticates with an OAuth token with a special scope), it would get back the scopes for any resource sets that it has registered for the current user. This downscoping should all be pluggable, too.

I'm going to pull out this service for now and see if there's a better way to handle it more generically. I'd like to work with you to make sure that you're still able to do what you need.

We don't know if the UMA branch is going to be merged into the next production 1.2 branch or if it's going to be targeted for 1.3 yet, but either way I want to make sure this doesn't step on your use case. Also please let me know if this requirement has is OBE.

alimfeld:
Thanks for the heads up! Allowing clients to introspect any token but only get back the scopes that it's registered for would work just fine for our use case.

PR #704, opened 2014-10-12
Add a useNonce attribute into ServerConfiguration to indicate if the IdP accepts the nonce value into its requests.

rafaelweingartner:
Changes requested on pull request 682 done.
Again, I am sorry for the trouble of opening a new pull request


jricher:
Just a note, this code has now been reverted from the main library as Google is now spec compliant and this hack is no longer needed.

PR #745, opened 2014-12-24
add @Transient to function getAuthorizedGrantTypes() so it does not pers...

cbelhaffef:
...ist


jricher:
Rebased and merged in e801505 and 7d649e5, thanks!

PR #763, opened 2015-02-11
Support for i18n of user interface

fjollberg:
Support for localised user interface with translations for Swedish.

Would it be possible to merge something like this?

The default locale resolver is currently a fixed locale resolver configured to use English, which can be overridden in overlays to fit into the locale resolve strategy at your site. There is currently no user interface added to select locale.

oic

Optimisations that can be made:
* Remove localised strings from HTML templates to decrease size.
* Use JSP for templates and use Spring for translations to the extent possible and cache templates
  in reverse proxy or similar in order to decrease the amount of JavaScript translations.


jricher:
This looks fantastic! We'll review this and look to pull it in to 1.2, thank you very much!

jricher:
I changed the configuration to use the existing ConfigurationPropertiesBean instead of a separate configuration file; this new resolver could be extended/replaced to handle per-user locales or other runtime configurations if desired.

The only thing I don't like about this approach is that the server-side translations and javascript translations are in separate files and formats. I'm not sure how to collapse these without having one mechanism's resource automatically create the other, but I'm filing a new issue to explore that.

Above all, thank you! This is great!

PR #779, opened 2015-03-08
Replaced DateUtil DateFormatter setup in MITREidDataServiceSupport

woltere:
Created Support class for MITREidDataService classes setting up the ```DateFormatter```

```fixObjectReferences()``` could also be moved to ```MITREidDataServiceSupport```, the 1_0 class has one line extra however 	```whitelistedSiteOldToNewIdMap.clear();```


jricher:
Merged, thank you. Do you think this is worth backporting to 1.1?

PR #803, opened 2015-04-07
Use the same "options" for token exchange that were needed for authorization request

w7tek:
With OpenAM server, several requests require an additional "realm" parameter. This 1-line change updates the token exchange request, to pass the same options that had been prepared for the initial authorization request to the OP.

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/openid/connect/client/OIDCAuthenticationFilter.java
+++ b/openid-connect-client/src/main/java/org/mitre/openid/connect/client/OIDCAuthenticationFilter.java
@@ -302,6 +302,9 @@ public class OIDCAuthenticationFilter extends AbstractAuthenticationProcessingFi
 		final RegisteredClient clientConfig = clients.getClientConfiguration(serverConfig);
 
 		MultiValueMap<String, String> form = new LinkedMultiValueMap<String, String>();
+
+		form.setAll(authOptions.getOptions(serverConfig, clientConfig, null));
+
 		form.add("grant_type", "authorization_code");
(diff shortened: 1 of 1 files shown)


w7tek:
@jricher What else do I need to do to evangelize for acceptance of this PR? Should I create an issue? The problem has to do with my org's implementation of OpenAM as OC requiring the "realm" parameter on certain requests, but I think that this one-line solution to my problem is general enough that it may help others. Certainly, it was the least invasive change I could find that solved my problem. There simply aren't enough extension points on the token-exchange code path in the filter.

jricher:
A pull request is fine, you don't need a separate issue as well. I've gotten the notifications, but I'm away for two weeks and haven't got the time to review and incorporate it. Thanks for your patience.

w7tek:
@jricher thanks for the heads-up. I can be patient :)

jricher:
I've reviewed your pull request and I think that this would be better handled by a separate Map being returned from the options object, effectively splitting it into getAuthorizationOptions and getTokenOptions or something like that.

w7tek:
what would you guess would be the timeline for implementing the change you're proposing instead of this?

w7tek:
> ...I think that this would be better handled by...

Could that be scheduled as a separate enhancement?
This simple PR fixes a specific case that is already currently broken. Are there cases that this PR breaks, motivating your proposed alternative?

jricher:
We don't have a timeline for this specific fix but if you're already running off of the development version of the project, I would suggest just running your fork until this issue has been closed.

PR #818, opened 2015-05-22
Update topbar.tag

### LINKED ISSUE #816: Small bug in topbar when family name is given
openid-connect-server-webapp/src/main/webapp/WEB-INF/tags/topbar.tag; Line 21:

```xml
<c:set var="longName" value="${ userInfo.givenName } {$ userInfo.familyName }" />
```
for the second variable, *{* and *$* are swapped. Should be:
```xml
<c:set var="longName" value="${ userInfo.givenName } ${ userInfo.familyName }" />
```

praseodym:
Fixes #816


jricher:
Good catch, thank you!

PR #819, opened 2015-05-23
Support for PostgreSQL

praseodym:


jricher:
Thanks, this is great. Are there many differences between the HSQL or MySQL table files and this one? I've got some changes in a branch that will be adding a stack of new tables to the database so I want to make sure I can put these all in parallel with each other.

praseodym:
I started with the HSQL file and only need to replace the generated id columns. In fact this is the entire diff between the HSQL and PostgreSQL files: https://gist.github.com/praseodym/86c04f7ed87aaa3d92e7

jricher:
Sounds good, that's a very similar change between the HSQL and MySQL versions, too.

jricher:
Rebased and merged in 742ceea18283478c8958111b0f5eaeda8ef18fcb

praseodym:
Cheers!

PR #820, opened 2015-05-23
Java 8 language level

praseodym:
I'm not sure if there are any good reasons to stay on Java 7 (which gets no more public updates), but the Java 8 language features are certainly worth upgrading for.

If this is accepted, I'll create a new PR to migrate some code to use Java 8 language features.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,2 @@
 language: java
+jdk: oraclejdk8
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
 	</mailingLists>
 
 	<properties>
-		<java-version>1.7</java-version>
+		<java-version>1.8</java-version>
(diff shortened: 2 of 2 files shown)


jricher:
We want to keep the compile target Java 7 for the 1.2 release for widest compatibility (some locked-down enterprise setups don't have as much of a choice), and we'll move to Java 8 (or 9) for the 1.3 release afterward. Considering that the production release targets 1.6 right now, this is a move in the right direction. :)

praseodym:
Sounds fair, maybe move this to a 1.3 milestone then?

jricher:
This will be targeted for 1.3, once the 1.3 branch is created.

jricher:
Accepted and applied in 2d495bd0bdbe4a84156fe624f5e94ee7476f6a78

PR #822, opened 2015-05-24
Add X-Frame-Options header

### LINKED ISSUE #762: Consider X-Frame-Options header to prevent clickjacking
The authorize screen in particular might want to  include a  X-Frame-Options header. Perhaps `SAMEORIGIN` or `DENY` would be a sane default value?

Details: https://tools.ietf.org/html/rfc6819#section-5.2.2.6

praseodym:
Fixes #762

I've also removed the HTTP security configuration for `/login**` because it seems unnecessary. I'd be happy to add it back if it makes sense.

### DIFF
--- a/openid-connect-server-webapp/src/main/webapp/WEB-INF/user-context.xml
+++ b/openid-connect-server-webapp/src/main/webapp/WEB-INF/user-context.xml
@@ -48,6 +48,9 @@
 		<security:logout logout-url="/logout" />
 		<security:anonymous />
 		<security:expression-handler ref="oauthWebExpressionHandler" />
+		<security:headers>
+			<security:frame-options policy="DENY" />
+		</security:headers>
 	</security:http>	
 
 </beans>


jricher:
Thanks, I'll look that over but it seems reasonable.

As I understand the workings of Spring Security, the HTTP block is necessary so that the /login page doesn't end up in a loop, where the login page requires login but you can't get there. I think this behavior depends on the kind of authentication filters you're using on the way in (most enterprise deployments of this don't use a username/password at the server), but we've had it this way for years with no noticeable trouble so I'm hesitant to pull it out. If you could submit that as a separate pull instead, I'll investigate it separately.

praseodym:
There is `<security:intercept-url pattern="/**" access="permitAll" />` in the bottom HTTP block, so the login page should never be behind authentication. I'll split it out to a different PR.

praseodym:
See #824 for the login block removal.

jricher:
Thanks for pulling it out separately. There's also a different entry point and session management on that page with this setup, which I recall causing problems without it. Might be redundant still in this basic setup though.

jricher:
Rebased and merged in d583499a07e16c7f067258be91d0c6a47f23b866

PR #823, opened 2015-05-24
Fix blacklist/whitelist for DynamicRegistrationClientConfigurationService

praseodym:
`ClientConfigurationService#getClientConfiguration` has ServerConfiguration as parameter, not String


jricher:
Good catch, thank you! We should have a unit test for this case (looks like we don't yet).

jricher:
Rebased and merged in effe955
(would love to get a unit test for this case if you get a chance)

PR #829, opened 2015-05-29
Fix inspections

praseodym:
General cleanup: fixes many code style warnings/notices as reported by IntelliJ IDEA's inspections.

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
+++ b/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
@@ -83,7 +83,7 @@ public class IntrospectingTokenService implements ResourceServerTokenServices {
 		}
 	}
 
-	private Map<String, TokenCacheObject> authCache = new HashMap<String, TokenCacheObject>();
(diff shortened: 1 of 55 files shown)


jricher:
Thanks for this patch. We can take some of these, but probably not all of them. We also need to update our checkstyle settings to scan for Java 1.7 paradigms, too.

jricher:
Rebased and patched, applied as 9e74e40453f0a6c74cc0827ea85505c1c56a1746 13f5e4f8a6aa5d457ef2aa2d1739c046e55a0abb 6dc2b2cb5ede2803c9f5461e08e97e0c4f4fed5f

praseodym:
Cheers! Any reason why the redundant initialisers commit has not been merged?

jricher:
Consistency in style. I'd rather have all variables that are defined on their own be initialized explicitly, instead of only when Java requires it. It's a habit I got into a long time ago and empty initializers look funny to me in Java now. 

I also pulled out a couple places that had collapsed an if/else into a boolean block because I believe the former to be more readable and understandable at a glance (at least in these specific cases).

Thanks very much for the patch!

praseodym:
That's fair, code style is personal preference. Thanks for merging!

PR #834, opened 2015-06-02
null check for expiration during introspection

wikkim:
Clients that are allowed to have tokens that dont expire can get tokens that dont have expiration dates. This makes the introspection service blow up on those tokens with NullExceptions.

Not sure if its appropriate to assume that null expirations means token hasn't expired or doesn't expire, but here's a pull request that does that.


jricher:
Thanks, that's worthwhile. I'll pull this in to the main branch (and probably backport to 1.1). 

This is @srmoore code, fwiw.

jricher:
Rebased and applied as 2d5e53c0f29b3bdcefaffbc0a311176d9f0e1395

wikkim:
I think you got the wrong commit hash :P 
This is the one 54fbf0d0ac3e7e1fee016c42a45955b08b85c2ae

jricher:
You are correct, had the wrong hash in my paste buffer; comment above corrected for whoever comes here later.

PR #846, opened 2015-06-18
Sync with HSQLDB version

lomori:
Few fields were missing, e.g. user_info::src, causing an application crash with MySQL.


praseodym:
The PostgeSQL definition is also missing the `user_info.src` column.

jricher:
Thank you. Development is usually done against the HSQL version and the MySQL (and now PSQL) versions are eventually sync'd.

jricher:
Rebased and merged as 84b595ea6f04a8d33bbe986b3bd36708521747f3, thanks

PR #872, opened 2015-07-22
use default system scopes if none specified when creating a new client

almalki:
The API documentation says that:
> If scope is omitted or null, all system scopes marked as "default" will be assigned to the client

However, I faced a bug where it is not doing this. I am not really sure where this logic (to use defaults) should be put, but I though the DefaultOAuth2ClientDetailsEntityService is the right place.

Am I missing something here? Is it really implemented before? If not, do you think this is the right place or should it be in ClientAPI.apiAddClient?

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/oauth2/service/impl/DefaultOAuth2ClientDetailsEntityService.java
+++ b/openid-connect-server/src/main/java/org/mitre/oauth2/service/impl/DefaultOAuth2ClientDetailsEntityService.java
@@ -135,6 +135,10 @@ public class DefaultOAuth2ClientDetailsEntityService implements ClientDetailsEnt
 		// check the sector URI
 		checkSectorIdentifierUri(client);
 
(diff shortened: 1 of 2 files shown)


jricher:
It happens here for Dynamically registered clients:

https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-server/src/main/java/org/mitre/openid/connect/web/DynamicClientRegistrationEndpoint.java#L378

When working through the Admin UI, the javascript app puts the default scopes in by default:

https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-server-webapp/src/main/webapp/resources/js/admin.js#L628

But it does look like the API when used directly doesn't actually enforce this. The problem here is that an *empty* scope list is fine and should be left empty, but a *null* or *missing* scope list needs to be handled appropriately. I don't think this is quite the way to handle it (given how other parts of the system handle scope processing), but thanks for bringing up the issue.

PR #874, opened 2015-07-24
Fixed up ApprovedSite AccessToken OneToMany performance issue.

httpants:
This patch significantly improves performance and I also believe it fixes the duplicate token defect that can happen if the cleanup task runs and has work to do (ie. there are expired tokens) at the same time as other threads are attempting to create a token for the same site.

### DIFF
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -62,12 +62,6 @@
       		<groupId>org.springframework.security.oauth</groupId>
  		 	<version>2.0.0.M2</version>
       		<artifactId>spring-security-oauth2</artifactId>
-            <exclusions>
-               <exclusion>
-                  <groupId>org.springframework.security</groupId>
-                  <artifactId>spring-security-config</artifactId>
-               </exclusion>
-            </exclusions>
 		</dependency>
         <dependency>
         	<groupId>com.nimbusds</groupId>
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -62,6 +62,12 @@
       		<groupId>org.springframework.security.oauth</groupId>
  		 	<version>2.0.0.M2</version>
       		<artifactId>spring-security-oauth2</artifactId>
+            <exclusions>
+               <exclusion>
+                  <groupId>org.springframework.security</groupId>
+                  <artifactId>spring-security-config</artifactId>
+               </exclusion>
+            </exclusions>
 		</dependency>
         <dependency>
         	<groupId>com.nimbusds</groupId>
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_0.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_0.java
@@ -835,7 +835,7 @@ public class MITREidDataService_1_0 extends MITREidDataService_1_X {
             }
             Long newGrantId = grantOldToNewIdMap.get(oldGrantId);
             ApprovedSite site = approvedSiteRepository.getById(newGrantId);
-            site.setApprovedAccessTokens(tokens);
+//            site.setApprovedAccessTokens(tokens);
             approvedSiteRepository.save(site);
         }
         accessTokenOldToNewIdMap.clear();
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_0.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_0.java
@@ -209,6 +209,7 @@ public class MITREidDataService_1_0 extends MITREidDataService_1_X {
     private Map<Long, String> accessTokenToClientRefs = new HashMap<Long, String>();
     private Map<Long, Long> accessTokenToAuthHolderRefs = new HashMap<Long, Long>();
     private Map<Long, Long> accessTokenToRefreshTokenRefs = new HashMap<Long, Long>();
+    private Map<Long, Long> accessTokenToApprovedSitesRefs = new HashMap<Long, Long>();
     private Map<Long, Long> accessTokenToIdTokenRefs = new HashMap<Long, Long>();
     private Map<Long, Long> accessTokenOldToNewIdMap = new HashMap<Long, Long>();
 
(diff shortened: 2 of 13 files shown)


jricher:
Thanks for digging this up. This is a fairly deep change so we're going to have to test out a lot of stuff to make sure it's not breaking anything, including the site management UI. I'm inclined to push this to 1.2 instead, but if we can make it work on this version without requiring an upgrade/export then it might be ok.

openid-connect-common/pom.xml
```diff
@@ -62,6 +62,12 @@
       		<groupId>org.springframework.security.oauth</groupId>
  		 	<version>2.0.0.M2</version>
       		<artifactId>spring-security-oauth2</artifactId>
+            <exclusions>
+               <exclusion>
+                  <groupId>org.springframework.security</groupId>
+                  <artifactId>spring-security-config</artifactId>
+               </exclusion>
+            </exclusions>
```
jricher: How is this exclusion related to this pull request?

openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_0.java
```diff
@@ -835,7 +835,7 @@ private void fixObjectReferences() {
             }
             Long newGrantId = grantOldToNewIdMap.get(oldGrantId);
             ApprovedSite site = approvedSiteRepository.getById(newGrantId);
-            site.setApprovedAccessTokens(tokens);
+//            site.setApprovedAccessTokens(tokens);
```
jricher: This looks like it would drop the approved site -> token binding during import from a 1.0 data export. There needs to be code to save this binding and put it in the right place.

openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_1.java
```diff
@@ -318,13 +318,13 @@ private void writeGrants(JsonWriter writer) throws IOException {
...
-            }
-            writer.endArray();
+//            Set<OAuth2AccessTokenEntity> tokens = site.getApprovedAccessTokens();
+//            writer.name("approvedAccessTokens");
+//            writer.beginArray();
+//            for (OAuth2AccessTokenEntity token : tokens) {
+//                writer.value(token.getId());
+//            }
+//            writer.endArray();
             writer.endObject();
```
jricher: Again, just like the above, this is dropping data on the floor without replacing it. You still need to export the approved->token links, and you can't actually change the export format like this. (The export can have fields added but you can't remove fields like this without breaking version compatibility.)

openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/MITREidDataService_1_1.java
```diff
@@ -1218,7 +1218,7 @@ private void fixObjectReferences() {
             }
             Long newGrantId = grantOldToNewIdMap.get(oldGrantId);
             ApprovedSite site = approvedSiteRepository.getById(newGrantId);
-            site.setApprovedAccessTokens(tokens);
+//            site.setApprovedAccessTokens(tokens);
```
jricher: Same as the 1.0 note: you need to link up the approved sites and tokens appropriately.

openid-connect-common/pom.xml
httpants: It's not related to the bug fix but it prevents 2 versions of spring-security-config appearing in my WEB-INF/lib.

jricher: OK, but please submit that as a separate issue/pull so as not to conflate the issues.

openid-connect-server/src/test/java/org/mitre/openid/connect/service/impl/TestMITREidDataService_1_1.java
```diff
@@ -1207,16 +1207,16 @@ public void testExportGrants() throws IOException {
...
-                    assertTrue(compare.getApprovedAccessTokens() == null || compare.getApprovedAccessTokens().isEmpty());
-                } else {
-                    assertNotNull(compare.getApprovedAccessTokens());
-                    Set<String> tokenIds = new HashSet<String>();
-                    for(OAuth2AccessTokenEntity entity : compare.getApprovedAccessTokens()) {
-                        tokenIds.add(entity.getId().toString());
-                    }
-                    assertThat(jsonArrayToStringSet(site.getAsJsonArray("approvedAccessTokens")), equalTo(tokenIds));
-                }
+//                if (site.get("approvedAccessTokens").isJsonNull() || site.getAsJsonArray("approvedAccessTokens") == null) {
```
jricher: It's a red flag to me that these unit tests were all commented out and not replaced with equivalents.

httpants: Ok this is actually the first pull request/patch i've submitted so may have assumed that i could submit something that was almost but not quite complete.  With the issues you've raised, what's the process to get them resolved?  Do i need to submit another pull request?

jricher: I definitely appreciate you taking the time to look into the issue and submitting a patch for it! You can actually just push additional commits to the same branch and the pull request will be updated with the latest commits. Changing a fundamental data model is going to touch a lot of different places throughout the code and will by no means be simple to make sure it's fully functional.

httpants: We're going live in about a week and i think if we could have a proper release with this change it will be alot more acceptable (to non technical upper level management) rather than us having our own custom patched version, so having a 1.1.17 release with this change would be nice.  It's does fix what i would classify as a major defect which i think is deserving of a special release in 1.1.x.

jricher: This patch breaks major, important functionality (the data import/export API which is the only way to do upgrades between major versions) and I have a feeling it breaks other pieces of functionality, such as the API and UI for managing the approved sites and associated tokens. If you want there to be a fixed version in a week, then please submit a fully functional patch that addresses these concerns and thoroughly tests all functionality within that time frame.

httpants:
I've updated the data import/export and test cases.

jricher:
This requested change has been re-implemented on the 1.3 branch. I'm not comfortable releasing this as a patch to the production versions.

PR #888, opened 2015-08-14
Updated commented-out datasource config for postgresql and mysql to use …

sdoxsee:
…HikariDataSource instead of Apache's BasicDataSource (since the Apache class isn't in the project by default)


jricher:
Oh, thanks! Looks like we missed that when we switched to Hiraki.

jricher:
Merged in 8d0355a

PR #890, opened 2015-08-20
OIDCAuthenticationFilter: Make authenticationSignerService optional so

BerndFrey:
it must not be provided in Spring config

OIDCAuthenticationProvider: Setter for UserInfoFetcher, so own
implementation can be wired

UserInfoFetcher: Call to DefaultUserInfo.fromJson moved to method, so it
can be overwritten by own implementation to use own UserInfo
implementation


jricher:
Thanks, both of these seem reasonable. I'd like to see some better error handling around the case if the signing service is null in the filter, though. Perhaps through a @PostConstruct check? Like if any clients have been configured to need it, and it's not there, then throw a reasonable error that will (hopefully) trigger before a null pointer is hit inside the filter.

jricher:
Rebased and applied in 9fe98e0

Issue #896, opened 2015-08-27
org.mitre.openid.connect.model.Address should be an interface


TonyFNZ:
UserInfo is an interface which is great.  This allows us to implement our own custom implementation without being tied to the default @Entity and @Table declarations.  

However UserInfo has a hard reference to the Address class which contains the @Entity and @Table declarations.  Address should be changed to an interface so that this can be more easily overridden by other parties.

Will try to submit a pull request for this when I can.

jricher:
Sounds reasonable to me. If this can be done without breaking the underlying data model (and existing database schema) then we can target this for inclusion in the current release stream, otherwise it will have to wait until 1.3 (which hasn't started yet).

jricher:
Closed by f43c94314c051d1ca68c916164c0200ad13acad1

PR #897, opened 2015-08-27
Issue #896: Change Address model to be an interface

TonyFNZ:
Simple code change to allow consuming projects to override the core library model more easily.  Doesn't break any of the underlying persistence layer.


jricher:
Rebased and merged as f43c94314c051d1ca68c916164c0200ad13acad1

Thanks!

PR #912, opened 2015-09-23
Addresses issue 910 by replacing update_time with update_at for JSON

iresprite:
objects.


jricher:
Rebased and merged at 719a7147352f8cbf46b82a33d2767c56c9961384

Had to pick up a couple of extra places that also used updated_time as well.

Thanks!

PR #916, opened 2015-09-29
Allow both flows authorization code and client credentials. 

coding-freedom:
Allow both flows authorization code and client credentials. This scenario might be found when the same client supports user authentication as well as service to service authentication. Such a client is trusted (whitelisted).

### DIFF
--- a/openid-connect-common/src/main/java/org/mitre/openid/connect/config/ConfigurationPropertiesBean.java
+++ b/openid-connect-common/src/main/java/org/mitre/openid/connect/config/ConfigurationPropertiesBean.java
@@ -56,7 +56,9 @@ public class ConfigurationPropertiesBean {
 
 	private Locale locale = Locale.ENGLISH; // we default to the english translation
 
-	public ConfigurationPropertiesBean() {
+    public boolean dualClient = false;
+
+    public ConfigurationPropertiesBean() {
 
 	}
 
@@ -168,4 +170,18 @@ public class ConfigurationPropertiesBean {
 	public void setLocale(Locale locale) {
 		this.locale = locale;
 	}
+
+    /**
+     * @return true if dual client is configured, otherwise false
+     */
+    public boolean isDualClient() {
+        return dualClient;
+    }
+
+    /**
+     * @param dualClient the dual client configuration
+     */
+    public void setDualClient(boolean dualClient) {
+        this.dualClient = dualClient;
+    }
 }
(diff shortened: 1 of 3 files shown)


jricher:
You can already enable this dual functionality with the existing client administration interface, which is the only way to whitelist a trusted client anyway. This functionality is restricted in the dynamic registration endpoint to prevent people from inadvertently creating a client with more access than they realized.

Is there a particular reason that you want this in the dynamic registration as well?

cosmin-cojocar:
The main reason to have this configuration in dynamic registration is that I use OIDC + OAuth 2 in a microservices architecture for a numerous number of services. In such a setup the lifecycle of a trusted OIDC client needs to be automated. Meaning that each service registers/deregisters itself dynamically with the oidc server. While the whitelisting is performed separately, as a DevOps task, for a predefined set of trusted clients.

jricher:
I don't like the idea of opening up that kind of functionality to dynamically registered clients, it seems to me that it would cause more trouble than anything by generally letting people cross client types like that. It's generally bad practice because the interactive and non-interactive clients in OAuth are generally for very different things.

For your use case, I can see two solutions that don't require code changes: (1) you can register two separate client IDs for each service that needs different kinds of access, one for the auth code and one for the client credentials. This would let you keep using dynamic registration, but you need to manage two client IDs for your clients. I recommend this approach. (2) you can authorize the clients to access the admin API directly. This would let you keep the multiple client types setup, but at the cost of giving server-level admin access to a set of clients. I wouldn't recommend it, but it's there.

cosmin-cojocar:
My first implementation was similar with your first recommendation, but just I want to reduce complexity and have only one client for both flows.

jricher:
That's fair. Perhaps we can make that restriction optional or pluggable on the endpoint, but leave the self-service UI as-is to prevent unintentionally open registrations. Would that work for your use case, and would you be willing to give that a shot in a pull request update?

cosmin-cojocar:
I would really appreciate any feedback from you. Thanks

jricher:
You mean in addition to the feedback already given? I'm not sure what you mean.

cosmin-cojocar:
I am not sure if this PR required additional changes besides  afaa15e in order to be merge. Just wanted to get your opinion.

jricher:
Oh, sorry, didn't see the additional commit. We'll review it and let you know. Thanks!

jricher:
Reviewed and merged in 7b34a666d9bf2c79d6a587c262d9798dffff29f4, thanks!

PR #929, opened 2015-10-08
Improve servlet startup speed

### LINKED ISSUE #928: `mvn jetty:run` not working
Running `mvn jetty:run` does not work; for some reason, the webapp never starts.

The problem here is that annotation scanning takes really long (2-3 minutes) when starting from the command line. Executing `mvn jetty:run-war -Dorg.eclipse.jetty.annotations.maxWait=180` gets around this. However, when deploying to Jetty from IntelliJ IDEA, everything is a lot faster. We need to figure out what causes this huge difference.

praseodym:
Update Jetty and configure `web.xml` for faster servlet start: 137580ms -> 20166ms for me.

Resolves #928.

### DIFF
--- a/openid-connect-server-webapp/pom.xml
+++ b/openid-connect-server-webapp/pom.xml
@@ -78,6 +78,7 @@
 				<groupId>org.eclipse.jetty</groupId>
 				<artifactId>jetty-maven-plugin</artifactId>
 				<configuration>
+					<war>${project.build.directory}/openid-connect-server-webapp.war</war>
 					<webAppConfig>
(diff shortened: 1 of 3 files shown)


jricher:
I didn't notice any speed differences, but it doesn't hurt my setup either. Merged in 301802abd3d8c57cb7eb7cb03bcbbc38ff889e45

praseodym:
For me it only affected Jetty when started from `mvn jetty:run`, Jetty launched from IntellIJ IDEA didn't have any problems. Nevertheless, it might benefit some production deployments as well.

PR #930, opened 2015-10-08
Improve error message

praseodym:
Improve error message for errors with a status code:

Before:
before

After:
after

These messages are usually swallowed by the logger (only visible on DEBUG level), so that's an extra argument for including them on the error page.


jricher:
Thanks, merged in 7871ee0f2603b98131c618a2729052b2da40c89f

PR #932, opened 2015-10-08
Remove legacy CSRF protection for approve page

### LINKED ISSUE #931: Approval fails with CSRF token error
Approving a new application (e.g. simple-web-app) in version 1.2.1 fails with a CSRF token error:
```
DEBUG: org.springframework.security.web.csrf.CsrfFilter - Invalid CSRF token found for http://localhost:8080/openid-connect-server-webapp/authorize]
```

This is caused by `approve.jsp` relying on 'proprietary' CSRF protection, even though the Spring Security filter is active on the `/authorize` endpoint.

praseodym:
Instead, we rely on the Spring Security CSRF protection, like we already do for the login page. Additionally, we remove the authentication check in`isApproved`, because this is already done by Spring Security (and if not, we have bigger problems to worry about).

Resolves #931.

Because this issue impacts core behaviour, I'd think this would be sufficient reason for a release of 1.2.2.

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/request/ConnectRequestParameters.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/request/ConnectRequestParameters.java
@@ -38,7 +38,6 @@ public interface ConnectRequestParameters {
 	public String PROMPT_SEPARATOR = " ";
 
 	// extensions
-	public String CSRF = "csrf";
 	public String APPROVED_SITE = "approved_site";
 
 	// responses
--- a/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/approve.jsp
+++ b/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/approve.jsp
@@ -260,8 +260,8 @@
 				</h3>
                 <spring:message code="approve.label.authorize" var="authorize_label"/>
                 <spring:message code="approve.label.deny" var="deny_label"/>
-				<input id="user_oauth_approval" name="user_oauth_approval" value="true" type="hidden" /> 
-				<input name="csrf" value="${ csrf }" type="hidden" />
+				<input id="user_oauth_approval" name="user_oauth_approval" value="true" type="hidden" />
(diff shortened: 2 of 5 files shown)


openid-connect-server/src/main/java/org/mitre/openid/connect/request/ConnectRequestParameters.java
```diff
@@ -18,31 +18,30 @@
...
 	// extensions
-	public String CSRF = "csrf";
-	public String APPROVED_SITE = "approved_site";
+	String APPROVED_SITE = "approved_site";
 
 	// responses
-	public String ERROR = "error";
-	public String LOGIN_REQUIRED = "login_required";
+	String ERROR = "error";
+	String LOGIN_REQUIRED = "login_required";
```
praseodym: The `public` modifiers here were removed by IntelliJ IDEA. According to the [JLS (§9.4)](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.4),
> Every method declaration in the body of an interface is implicitly public (§6.6). It is permitted, but discouraged as a matter of style, to redundantly specify the public modifier for a method declaration in an interface.

jricher:
Please revert the removal of "public" from the interface definition as this makes it inconsistent with the rest of the project's style. I'm aware of that style suggestion but it is not applied here. 

Thanks.

praseodym:
Done.

sdoxsee:
Was tripped up by this for the last day :P

This fix sounds great. Would it be possible to merge this and re-release @jricher ? I can't get 1.2.1 working without manually doing all this in my overlay project. Thanks @praseodym!

sdoxsee:
actually, I'm not in production with this yet so I can get around with other testing for now by taking away the csrf protection in user-context.xml for now.

jricher:
Thanks! I can possibly do the pull and release next week. I was having trouble with the integration test environment just before the last release which is how it slipped by. Oops.

sdoxsee:
Thanks for being all over this! Overlay updated with 1.2.2. Cheers!

PR #935, opened 2015-10-11
Fix ID token icon description

praseodym:
The ID token icon currently has the description from the refresh token icon.


jricher:
Thanks, merged in 58543ac9c4bd53ee442ddb36636db5fa26e2c60e

PR #937, opened 2015-10-12
Spring Maven BOM dependency + remove commons-logging

praseodym:
The [Maven "bill of materials" dependency](http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#overview-maven-bom) ensures that all Spring dependencies use the same version, without having to specify them all.

Furthermore, we remove all transitive commons-logging dependencies that are currently still present. Unfortunately Maven does not have a way to do this globally.

### DIFF
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -31,6 +31,12 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
+			<exclusions>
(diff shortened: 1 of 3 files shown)


jricher:
This looks promising, thanks!

praseodym:
Now it works :sparkles:

jricher:
Rebased and merged in d337e14de3f83fdc954954b4a24647294dc4a492

PR #943, opened 2015-10-15
Upgrade to Spring Framework 4.1.8

praseodym:
http://spring.io/blog/2015/10/15/spring-framework-4-2-2-4-1-8-and-3-2-15-available-now

The Spring Framework 4.2 series have also been stable for a while now, but maybe that's more suited for a 1.3 release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -259,7 +259,7 @@
 			<dependency>
 				<groupId>org.springframework</groupId>
 				<artifactId>spring-framework-bom</artifactId>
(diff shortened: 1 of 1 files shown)


jricher:
Thanks! And yes, changing versions of Spring/SpringSecurity beyond patch releases is more of a breaking change (unfortunately) so we're sticking with the current streams until the next release.

jricher:
Merged in Commit: abff7421c1872f4b5c53d724ea389b4a87052ad8, thanks!

PR #946, opened 2015-10-20
Make FILTER_PROCESSES_URL public

praseodym:
Currently hardcoded in the filter and the client's Spring Security config; would be nicer to reference the value instead.

https://github.com/mitreid-connect/simple-web-app/blob/c5e70ebd5c28de2cce81d106435ae11925424a5f/src/main/webapp/WEB-INF/spring/appServlet/servlet-context.xml#L54


jricher:
Merged in Commit: a3d01727f96da2e36a4820c82a62020fbdcd9538, thanks!

PR #950, opened 2015-10-25
Filter only `*.tag` files

praseodym:
Instead of excluding all binary files, just filter `*.tag` files.
http://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html


jricher:
The excludes-filter grew over time as we discovered more things were breaking, so this makes a lot more sense. We don't need to filter JSPs as well?

praseodym:
From what I could see filtering is only used for `${project.version}` in `copyright.tag`, I couldn't find anything in JSPs.

jricher:
Merged in 4c1e6866ce73bfc9864e1870d1b5c4f1ff18a5f3 with filtering added to JSPs as well, thanks!

PR #951, opened 2015-10-25
Improve state handling in handleAuthorizationCodeResponse

### LINKED ISSUE #949: Unexpected exception in client when session is lost mid-authentication
Steps to reproduce are as follows:

1. Have OIDC client application running on localhost
2. Go to 127.0.0.1:8080, session is now bound to this address
3. Login using OIDC, redirects back to localhost:8080
4. Session is empty (no cookie on "localhost:8080") and unexpected exception occurs

The exception is a NPE or whitelist error in DynamicServerConfigurationService.

We found the root cause to be in [the OIDCAuthenticationFilter](https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/5f24685f1786a4556ba8e101e1ae1028d1e89ec7/openid-connect-client/src/main/java/org/mitre/openid/connect/client/OIDCAuthenticationFilter.java#L285-L297): when the session does not contain the `state` or `issuer` attributes (i.e. they are null), the state check does not fail and a null `issuer` is passed down to the `ServerConfiguration` class and causes an NPE there.

We should ensure that the state check fails when the `state` attribute is not empty or null; can it disappear for a legitimate reason?

praseodym:
Fail fast when there is no state in session, e.g. because the session
cookie was removed.

Resolves #949

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/openid/connect/client/OIDCAuthenticationFilter.java
+++ b/openid-connect-client/src/main/java/org/mitre/openid/connect/client/OIDCAuthenticationFilter.java
@@ -286,11 +286,9 @@ public class OIDCAuthenticationFilter extends AbstractAuthenticationProcessingFi
 
 		// check for state, if it doesn't match we bail early
 		String storedState = getStoredState(session);
-		if (!Strings.isNullOrEmpty(storedState)) {
-			String state = request.getParameter("state");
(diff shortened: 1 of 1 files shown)


jricher:
Merged in 4f9ea0b474ae481fa20365744841a4c0db0c9a05, thanks!

Issue #553, opened 2014-02-28
Upgrade to Bootstrap 3


jricher:
Bootstrap 2 has reached EOL, we should try to upgrade and track with the newer version at some point. Probably post 1.1-branch though.

jricher:
Icons don't seem to translate well when locally served, this might keep us off Bootstrap3 and FontAwesome.

hansgru:
> Icons don't seem to translate well when locally served, this might keep us off Bootstrap3 and FontAwesome.

I don't understand what you mean here. We are using Boostrap3 and FontAwesome on in many projects (and upgraded them from 2 to 3 for Bootstrap and 3 to 4 for FontAwesome without problems)

jricher:
When I've tested this under Firefox on several platforms, I was not able to get the web font to load from the server's local domain. Everything runs fine when running from a CDN, but many of our deployments are on closed or restricted networks, so we need to pack in everything. We already have CORS headers set and the paths seem fine, so at the time I couldn't figure out what was causing the hiccup. Any advice would be appreciated!

hansgru:
FontAwesome  is using special font files for those icons (not images).

The difference can be in the encoding and mime types of the resources that come from CDN vs those who come from the local server. 

E.g. apps made with Spring (e.g. Grails) won't handle font files correctly out of the box so extra configuration is required for fonts files to be served correctly by a local server too (not just images and css as they usually do)

FireBug can help to pinpoint this by having the 2 versions of a page and see if the local vs CDN headers are similar regarding encoding, mime type but also the content (e.g. in case Spring Security is not letting the font files tru).

(we also enforce all our Tomcats to have in catalina.bat/sh ```JAVA_OPTS=-Dfile.encoding=UTF8``` to avoid further encoding mess)

jricher:
I've just checked the headers for the local and CDN versions of the .eot file, and they aren't identical but are fairly close and I can't see what's causing the issue.

Local:

```
    Status Code: 200 OK
    Access-Control-Allow-Origin: *
    Content-Length: 67119
    Content-Type: application/vnd.ms-fontobject
    Date: Thu, 12 Mar 2015 15:38:57 GMT
    Last-Modified: Thu, 12 Mar 2015 15:36:26 GMT
    Server: Apache-Coyote/1.1
```

CDN:

```
    Status Code: 200 OK
    Access-Control-Allow-Origin: *
    Cache-Control: max-age=31104000
[...]

PR #952, opened 2015-10-25
Align user profile claims

praseodym:
Especially helps when `<dd>` collapses due to an empty claim value.

Related to https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/issues/553#issuecomment-118527610


jricher:
Merged in Commit: f27673a5f5e6108766420b3c085e3e86b1dbd590, thanks!

Will we need to back this out during the Bootstrap 3 upgrade?

praseodym:
If Bootstrap 3 indeed fixes this problem (haven't actually tried) we probably should.

PR #953, opened 2015-10-25
Clean up ScopeClaimTranslationService

praseodym:
`getFieldNameForClaim` method is never used.


jricher:
I was on the fence about this one, but it seems straightforward to remove it. I think we put it in originally for parallelism thinking we might need it later, but never did.

Merged in Commit: dce80d488b7806fb43f27d6567a1d401883bf390, thanks!

PR #954, opened 2015-10-27
Fix JPA issues to allow using Hibernate

mtommila:
There are some issues, which prevent using Hibernate (instead of EclipseLink) as the JPA provider. Fixing these issues should allow using Hibernate. It would seem like some of the issues are JPA violations (i.e. bugs) but EclipseLink is lenient enough to handle them. Some things might be just Hibernate quirks but fixable nevertheless.

Using Hibernate is nice e.g. for better performance, more flexible 2nd level JPA caching, and JMX monitoring.

### DIFF
--- a/openid-connect-common/src/main/java/org/mitre/oauth2/model/AuthenticationHolderEntity.java
+++ b/openid-connect-common/src/main/java/org/mitre/oauth2/model/AuthenticationHolderEntity.java
@@ -66,7 +66,7 @@ public class AuthenticationHolderEntity {
 
 	private SavedUserAuthentication userAuth;
 
-	private Collection<? extends GrantedAuthority> authorities;
+	private Collection<GrantedAuthority> authorities;
 
 	private Set<String> resourceIds;
 
@@ -116,14 +116,14 @@ public class AuthenticationHolderEntity {
 
 		// pull apart the request and save its bits
 		OAuth2Request o2Request = authentication.getOAuth2Request();
-		setAuthorities(o2Request.getAuthorities());
+		setAuthorities(o2Request.getAuthorities() == null ? null : new HashSet<>(o2Request.getAuthorities()));
 		setClientId(o2Request.getClientId());
-		setExtensions(o2Request.getExtensions());
+		setExtensions(o2Request.getExtensions() == null ? null : new HashMap<>(o2Request.getExtensions()));
 		setRedirectUri(o2Request.getRedirectUri());
-		setRequestParameters(o2Request.getRequestParameters());
-		setResourceIds(o2Request.getResourceIds());
-		setResponseTypes(o2Request.getResponseTypes());
-		setScope(o2Request.getScope());
+		setRequestParameters(o2Request.getRequestParameters() == null ? null : new HashMap<>(o2Request.getRequestParameters()));
+		setResourceIds(o2Request.getResourceIds() == null ? null : new HashSet<>(o2Request.getResourceIds()));
+		setResponseTypes(o2Request.getResponseTypes() == null ? null : new HashSet<>(o2Request.getResponseTypes()));
+		setScope(o2Request.getScope() == null ? null : new HashSet<>(o2Request.getScope()));
 		setApproved(o2Request.isApproved());
 
 		if (authentication.getUserAuthentication() != null) {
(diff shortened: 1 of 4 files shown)


jricher:
The wrappers for new maps and sets are required to ensure writable collections in the managed objects, we need to keep those in. Do they break hibernate?

mtommila:
Hibernate injects its own Map and Set implementations to collection properties. If in the setter method another collection implementation is used and the elements are copied to it from the method argument collection, this breaks Hibernate. Apparently it does not allow reading the collection while the entity is still being initialized.

I was assuming that the problematic cases are:

AuthenticationHolderEntity.setAuthentication(OAuth2Authentication)

and

new SavedUserAuthentication(Authentication)

But these are handled in this pull request.

When initializing an AuthenticationHolderEntity from an OAuth2Authentication, we don't know if the collections from the OAuth2Authentication are writable so we must ensure writable collections by using a new HashSet.

Also SavedUserAuthentication obviously needs to use a writable Set as the Set from Authentication.getAuthorities() will be immutable.

But are there also other cases where an immutable collection is put to an entity with a setter method, and then the code expects that the collection from the corresponding getter is mutable, immediately after the setter call (not managing the object with JPA in between)?

Generally I think Hibernate will allow any collection implementation to be used in an entity's property, it just breaks when a collection property is read during the entity's initialization in the JPA provider. When you persist the object to JPA then I suppose it will always replace the collection implementation with its own.

I have tested the code in our application and it works, so in that sense I am not worried.

jricher:
Ok, I'll test these out on eclipse link to see how it works out on different db platforms. If it doesn't break existing systems we should be fine to pull it in, but I just remember having to put these wrappers in place for similar to the reasons you've mentioned.

mtommila:
Thanks! One more consideration might be that the JPA provider can (and will) try to optimize the collection handling. So it plugs in its own collection implementation to a managed entity, to track which elements were retrieved from and put into the collection. This makes it easier (and more efficient) for the JPA implementation to track, which elements are "dirty" and what it needs to actually persist to the database. If the collection is just a standard HashMap or ArrayList then it's more difficult for the JPA provider to figure out what are the changes made to the collections and the elements in them.

gleeworks:
We are looking into using hibernate as well, and ran into this issue as well.  I stumbled upon this pull request thread, and the changes seem to work for me locally applied to 1.2.7-SNAPSHOT with some preliminary testing with MySQL.  Since it's been awhile, I just wanted to +1 the desire for this fix in an eventual release, if this is still agreeable.  Thanks!

jricher:
I've pulled these into 1.3 as 4f4c8de1c8f662ec74e4b4c99ec06a23c0cb66c5 (still pending merge in a dev branch) but I had to change one item to make it work: the "attributeName" field on the annotation for the extensions map in the authentication holder class does not work with the EclipseLink JPA implementation. If this is required for Hibernate, we'll need to find a workaround that doesn't break EclipseLink.

leonard84:
@mtommila / @jricher wouldn't it be better to keep the wrappers and use `@Access(AccessType.FIELD)` instead? If you move the JPA annotations to the fields instead of the getters then it should use field access by default. 

The old version enforced set semantics,
```java
public void setAuthorities(Collection<? extends GrantedAuthority> authorities) {
		if (authorities != null) {
			this.authorities = new HashSet<>(authorities);
		} else {
			this.authorities = null;
		}
	}
```
The new code will accept a list just as fine, so you can get duplicates.

Furthermore, collection fields should never be null, the persistence layer will treat an null the same as an empty collection and when it is a managed entity it will insert an empty collection. So to unify access the collection should be initialized. This way the code would change to something like this.

```java
	@ElementCollection(fetch = FetchType.EAGER)
	@CollectionTable(
			name="authentication_holder_authority",
			joinColumns=@JoinColumn(name="owner_id")
			)
	@Convert(converter = SimpleGrantedAuthorityStringConverter.class)
	@Column(name="authority")
	private Collection<? extends GrantedAuthority> authorities = new HashSet<>();

	public void setAuthorities(Collection<? extends GrantedAuthority> authorities) {
			this.authorities.clear();
			this.authorities.addAll(authorities);
	}

	public Collection<? extends GrantedAuthority> getAuthorities() {
		return authorities;
	}
```
So the client code can use `getAuthorities()` without having to worry about it being null.

jricher:
The reason we're not using field-style access types is that you can't mix field and method style annotations in JPA -- at least not when the project started, I haven't looked to see if that changed -- and we've got several cases where the data isn't a simple mapping to a local variable.

leonard84:
@jricher do you mean you can't mix the access for a single class or for all classes in the persistence context?
However, both is possible now: http://thebrainyprofessionals.com/2013/05/26/mixed-access-types-in-jpa/

PR #970, opened 2015-11-13
Bugfix for multiple client redirects in UUIDPairwiseIdentiferService

oharsta:
When a client has multiple redirect URIs, no configured
sectorIdentifierUri and the subject type is PAIRWISE then
the UUIDPairwiseIdentiferService#getIdentifier should not
fail, but instead use the first redirect URI for calculating
the sector identifier. Fixes#969

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ bin
 /target
 .springBeans
 nb-configuration.xml
+.java-version
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/UUIDPairwiseIdentiferService.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/service/impl/UUIDPairwiseIdentiferService.java
@@ -31,6 +31,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.util.CollectionUtils;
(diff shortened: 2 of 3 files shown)


jricher:
According to http://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg if the client has more than one redirect URI then it needs to have a sector identifier. The `isEmpty` check should actually be a check for a single value and keep the `isOnlyElement` call within. If there are multiple redirect URIs and no sector URI then the pairwise calculation should fail.

oharsta:
Then the GUI constraints should be updated according spec I guess. I'll amend the pull request. With "fail" you mean return null or throw SomeKindOfRuntimeException. I think the latter would make more sense.

jricher:
If an exception is thrown then it needs to be caught and handled by something. Right now the signaling is set up to look for the null return in an error condition, so unless you want to also change the upstream handlers I'd stick with that.

praseodym:
I'd prefer a runtime exception over returning null, unless the upstream code handles these nulls nicely. If not, we'll be turning a potentially descriptive runtime exception into an NPE somewhere else.

jricher:
Checking the unit tests they were expecting an `InvalidArgumentException` anyway, so we'll do that. I'm not sure why that unit test wasn't triggered with current behavior.

oharsta:
So it was not a bug, but according spec. Still would be nice if the GUI would warn us if we create a client with multiple redirect URIs, no sector identifier and the subject type is PAIRWISE. But that is not a high prio for me.

Issue #891, opened 2015-08-21
Implementation of End Session Endpoint


selvarajkumarasamy:
Jricher : we have implemented "end session end point" based on the openid connect specification,but "check session end point" not implemented.
I have following queries in above implementation.
1.whether need to clear the token in database?
2.how to implement single app log out?

Give me details, does anybody implementing the same?.

Thanks 
Selvaraj K

jricher:
Yes, you should delete the ID token from the database. 

To implement single app logout, you need to prompt the user at the end session endpoint and also call all current RP's end session endpoints for that user with their tokens.

selvarajkumarasamy:
Jricher: Thanks for the update..we follow the same.

jricher:
Implemented in 1.3.1

Issue #856, opened 2015-07-02
Is the RP-Initiated Logout scenario already implemented?


ngriesser:
Setting the post_logout_redirect_uri in the database does not have any effect, calling the logout endpoint with a post_logout_redirect_uri parameter redirects to the sso server itself.

jricher:
The logout endpoint is not meant to be called externally. RP-initiated logout is not implemented yet.

ngriesser:
So to implement it the right way, we should add another logout endpoint that asks the user for consent and eventually calls the internal logout. Am I getting it right?

jricher:
Correct, and it should offer the users the option of logging out of the IdP and potentially sending logout messages to all other RPs. The internal logout URL really should have some CSRF protection on it anyway to prevent external callers from doing a drive-by logout.

PR #972, opened 2015-11-16
Add basic implementation of end session endpoint

sdoxsee:
This addresses the RP-initiated logout section of the session management spec (https://openid.net/specs/openid-connect-session-1_0.html) and has some gaps. I'm creating this PR in case other people want a starting point for implementing it themselves (rather than start with nothing).

### DIFF
--- a/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/logout.jsp
+++ b/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/logout.jsp
@@ -1,0 +1,38 @@
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<%@ taglib prefix="o" tagdir="/WEB-INF/tags"%>
+<%@ taglib prefix="security" uri="http://www.springframework.org/security/tags"%>
+<%@ taglib prefix="spring" uri="http://www.springframework.org/tags"%>
+
+<spring:message code="logout.title" var="title"/>
+<o:header title="${title}"/>
+<o:topbar pageName="Logout"/>
+<div class="container-fluid main">
+	<div class="row-fluid">
+		<div class="span1"></div>
+		<div class="span10">
+            <security:authorize access="hasRole('ROLE_USER')">
+				<div class="hero-unit">
+					<h2><spring:message code="logout.title"/></h2>
+					<p>
+						<spring:message code="logout.body"/>
+					</p>
+					<p>
+						<a href="" class="btn btn-primary btn-large endSessionLogoutLink"><spring:message code="topbar.logout"/></a>
+		                <form action="${ config.issuer }${ config.issuer.endsWith('/') ? '' : '/' }logout" method="POST" class="hidden" id="endSessionLogoutForm">
+							<input type="hidden" name="${_csrf.parameterName}" value="${_csrf.token}" />
+						</form>
+					</p>
+				</div>
+            </security:authorize>
+		</div>
+	</div>
+</div>
+<o:footer/>
+<script type="text/javascript">
+	$(document).ready(function() {
+		$('.endSessionLogoutLink').on('click', function(e) {
+			e.preventDefault();
+			$('#endSessionLogoutForm').submit();
+		});
+	});
+</script>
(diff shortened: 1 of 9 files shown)


sdoxsee:
Reference: #891 and #856

openid-connect-server/src/main/java/org/mitre/openid/connect/web/EndSessionEndpoint.java
```diff
@@ -0,0 +1,97 @@
...
+ *
+ */
+@Controller
+@PreAuthorize("hasRole('ROLE_USER')")
+public class EndSessionEndpoint {
+
+	public static final String URL = "endsession";
+	
+	@Autowired
+	EndSessionValidator endSessionValidator;
```
jricher: Please make all these member variables private.

openid-connect-server/src/main/java/org/mitre/openid/connect/web/EndSessionValidator.java
```diff
@@ -0,0 +1,78 @@
...
+		} else if (accessToken.getClient().getPostLogoutRedirectUris() == null ||
+				!accessToken.getClient().getPostLogoutRedirectUris().contains(postLogoutRedirectUri)) {
+			logger.info("Unregistered post_logout_redirect_uri: " + postLogoutRedirectUri);
+			return false;
+		} else if (accessToken.getAuthenticationHolder() == null ||
+				accessToken.getAuthenticationHolder().getUserAuth() == null ||
+				accessToken.getAuthenticationHolder().getUserAuth().getName() == null ||
+				auth == null ||
+				auth.getPrincipal() == null ||
+				!accessToken.getAuthenticationHolder().getUserAuth().getName().equals(auth.getPrincipal())) {
```
ajmesa9891: Shouldn't this be .equals(auth.getName())? You want to make sure the two names match, and even though the principal, which is an Object, could be a String with the name, it may not. Notice how the getName method in auth will return the principal (that your code uses directly) if it is not a UserDetails or Principal, so in your specific case it would not have any negative side-effects.

sdoxsee: Thanks @ajmesa9891. You make a good point about Object vs String. I'm not able to verify the correctness of this at the moment but if the getName() method indeed provides the username then we should go with that rather than getPrincipal(). Would you be willing to update the tests and change lines 43 and 44 accordingly?

openid-connect-server-webapp/src/main/webapp/WEB-INF/views/logout.jsp
```diff
@@ -0,0 +1,38 @@
...
+		<div class="span1"></div>
+		<div class="span10">
+            <security:authorize access="hasRole('ROLE_USER')">
+				<div class="hero-unit">
+					<h2><spring:message code="logout.title"/></h2>
+					<p>
+						<spring:message code="logout.body"/>
+					</p>
+					<p>
+						<a href="" class="btn btn-primary btn-large endSessionLogoutLink"><spring:message code="topbar.logout"/></a>
```
tbroyer: So you're using a link that links to "nowhere", that you style to make it look like a button, and that you script to prevent the navigation and instead submit a form. Wouldn't it be easier (and much cleaner) to just use a submit button in the form?

sdoxsee: @tbroyer you're probably right. I was blindly following what had been implemented for logout. e.g. https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-server-webapp/src/main/webapp/WEB-INF/tags/topbar.tag#L130

tbroyer: In the case of the linked logout page, it's because there are 2 "logoutLinks" (only one shown at a time depending on viewport size, relying on Bootstrap "responsiveness"), and instead of duplicating the form (and using submit buttons), the author chose to instead rely on JavaScript.

coderReview:
I have a doubt about the id token hint.
After the authorization code flow, to get the id token a request should be made to the token endpoint. Is this correct?  Since the server can retrieve the user id token, is id_token_hint a required parameter? 
Thanks

jricher:
Implemented in 1.3.1

Issue #982, opened 2015-12-04
Deleting refresh token corrupts access_token table


mtommila:
The access token entity has a unidirectional many-to-one association to refresh token. There are many ways to just delete refresh tokens, this can leave the data in the access_token table in a corrupted state. I.e. the refresh_token_id column contains a value that does not actually exist in the refresh_token table.

Maybe the refresh token entity should also have an association back to the access token entity. This way, when the refresh token is deleted, the link from access_token would be set to null.

Or, alternatively, this would have to be done manually when a refresh token is deleted (find access tokens that point to the refresh token, set the association to null, then delete the refresh token).

Database foreign key constraints would also be one way to enforce that this cannot happen. But there are no foreign key constraints.

jricher:
Where in the code can you delete a refresh token without its attendant access tokens? The token services should take care of this automatically. I'm happy to have it be double-checked though.

mtommila:
We had a data corruption although I'm not entirely sure how it happened. I suspect (since we had two servers running in a cluster) that the clean-up jobs happened to run in parallel. So both servers get a list of access tokens / refresh tokens / authentication holders / authorization codes to be cleaned up, and then start cleaning them up. Only it would happen so that both servers get the same list of entities to be deleted, and then have a race who happens to delete them. The other then fails as the token is already deleted by the other server. At a suitable point of time things end up so that only partial deletion occurs and data is left in a corrupted state.

DefaultOAuth2ProviderTokenService.clearExpiredTokens() removes access tokens, refresh tokens and authentication holders.

DefaultOAuth2AuthorizationCodeService.clearExpiredAuthorizationCodes() removes authorization codes.

We have since configured the clean-up to only run on one of the cluster nodes, and I'm not sure if the data corruption can occur anymore.

One suggestion to possibly make things more robust would be to use bulk deletes instead of a select and then looping through the results and deleting each entity individually.

Also we use Hibernate as the persistence provider instead of EclipseLink. There may be some assumptions that EclipseLink makes that do not hold in Hibernate.

jricher:
Yeah, if you have the cleanup tasks running on two servers like that then things can get weird. We don't do bulk deletes precisely to allow for this type of measured cleanup and deletion to occur.

PR #1015, opened 2016-02-04
Fix issue where under high load the token expiration job fails.  

strangeweaver:
At present under high load the token expiration task can fail. This is because a maximum of 1000 access tokens are cleared in a single run. Under high load the volume of access tokens can grow sufficiently large to cause the aggregate query used to remove duplicate access tokens to timeout, with continuing failures overall application performance degrades.

The removal of duplicate access tokens is necessary because:

1) Neither the application nor database enforce uniqueness on the access token value.
2) the application expects the access token value to be unique upon removal of an access token.

These two expectations are incompatible and lead to data inconsistency within the application. To fix this we can lookup the accessToken to be removed by the primary key column (not value) and remove it. The presumption here is that the intent of a method removeAccessToken passing an Entity. Is to remove the entity passed and only that entity.

Fixing the inconsistency between the insert and delete assertion then removes the need to clear the duplicate tokens as they will be removed eventually.

### DIFF
--- a/openid-connect-common/src/main/java/org/mitre/data/AbstractPageOperationTemplate.java
+++ b/openid-connect-common/src/main/java/org/mitre/data/AbstractPageOperationTemplate.java
@@ -1,0 +1,160 @@
+package org.mitre.data;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Abstract class for performing an operation on a potentially large
+ * number of items by paging through the items in discreet chunks.
+ *
+ * @param <T>  the type parameter
+ * @author Colm Smyth.
+ */
+public abstract class AbstractPageOperationTemplate<T> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AbstractPageOperationTemplate.class);
+
+    private static int DEFAULT_MAX_PAGES = 1000;
+    private static long DEFAULT_MAX_TIME_MILLIS = 600000L; //10 Minutes
+
+    /**
+     * int specifying the maximum number of
+     * pages which should be fetched before
+     * execution should terminate
+     */
+    private int maxPages;
+
+    /**
+     * long specifying the maximum execution time
+     * in milliseconds
+     */
+    private long maxTime;
+
+    /**
+     * boolean specifying whether or not Exceptions
+     * incurred performing the operation should be
+     * swallowed during execution default true.
+     */
+    private boolean swallowExceptions = true;
+
+
+    /**
+     * default constructor which sets the value of
+     * maxPages and maxTime to DEFAULT_MAX_PAGES and
+     * DEFAULT_MAX_TIME_MILLIS respectively
+     */
+    public AbstractPageOperationTemplate(){
+        this(DEFAULT_MAX_PAGES, DEFAULT_MAX_TIME_MILLIS);
+    }
+
+
+    /**
+     * Instantiates a new AbstractPageOperationTemplate with the
+     * given maxPages and maxTime
+     *
+     * @param maxPages the maximum number of pages to fetch.
+     * @param maxTime the maximum execution time.
+     */
+    public AbstractPageOperationTemplate(int maxPages, long maxTime){
+        this.maxPages = maxPages;
+        this.maxTime = maxTime;
+    }
+
+
+    /**
+     * Execute the operation on each member of a page of results
+     * retrieved through the fetch method. the method will execute
+     * until either the maxPages or maxTime limit is reached or until
+     * the fetch method returns no more results. Exceptions thrown
+     * performing the operation on the item will be swallowed if the
+     * swallowException (default true) field is set true.
+     */
+    public void execute(){
+        logger.info("Starting execution of paged operation. maximum time: " + maxTime
+                + " maximum pages: " + maxPages);
+
+        long startTime = System.currentTimeMillis();
+        long executionTime = 0;
+        int i = 0;
+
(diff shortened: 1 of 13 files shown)


jricher:
Interestingly, all of the mechanisms listed above (the 1000 token limit, fixing duplicate tokens) were put in for performance issues.

strangeweaver:
Yes I had noticed that . We are also working on the issue of the 1000 token limit where we use a limit but permit the task to reiterate until no records are returned or a max time limit is achieved, all of which will be configurable but will have sensible defaults. I'll commit once we have it fully prepped and tested.

praseodym:
Shouldn't it be possible to issue a `DELETE` query and let the database cascade the delete to whatever other rows are associated with these tokens?

strangeweaver:
Hi Praseodym,

I had considered a DELETE query also, but it doesn't seem to be consistent with the general approach taken in this software project. Deletes throughout the project are done by retrieving items from the entity manager before deleting them, I assume this is done

1) to avoid attempting to delete detached entity.
2) to ensure that that state of the first level cache is maintained (delete queries don't update the persistence context).

I assume this is a design decision in the product and have therefore crafted this fix to be consistent with the products current design (I seem to remember a discussion topic around why individual deletes were chosen over bulk deletes but haven't been able to dig it up)

Also even if a DELETE query were used, it should be a new method on the repository e.g. bulkDeleteAccessTokensByValue(String value).

While this pull request solves an performance issue, it is really more about correcting what I perceive to be inconsistencies in the API, 

namely token value uniqueness while likely is not guaranteed at creation, while token value uniqueness is required at deletion.

- A unique constraint is a possibility but is a discussion change for a major release.
- Another possible solution would have been to create a method removeAccessTokensByValue(String value) which would retrieve a collection of attached entities with the token value and then remove them (thereby maintaining the persistence context). 

 would contend, however, that this is a reasonable change regardless of performance issues. My expectation would still be that removeAccessToken(OAuth2AccessTokenEntity) should remove the exact OAuth2AccessTokenEntity passed in (i.e. by primary key) not by using the OAuth2AccessTokenEntity as criteria (i.e. by use value).

praseodym:
I agree that this at least fixes the currently inconsistent behaviour with regards to deleting by non-unique values.

For the next version (1.3) we should definitely consider a DELETE query. [EclipseLink supports this and can update the persistence context](https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#Delete_Queries), so I don't see real problems there.

praseodym:
Also, could you squash the two commits in this PR together?

strangeweaver:
Hi,

I tried squashing those two commits together but must admit I'm pretty new to git and don't fully understanding how the rebase mechanism works. It seems to have created 4 commits instead of reducing to one. Can you advise?

In regards to delete queries, I agree they could prove very useful for the token expiration job. Incidentally the EclipseLink you pasted is the one I was quoting. The two final statements are somewhat contradictory. The first one says

**The persistence context is not updated to reflect results of delete operations**. If you use a transaction-scoped persistence context, you should either execute the bulk operation in a transaction all by itself, or be the first operation in the transaction. **That is because any entity actively managed by the persistence context will remain unaware of the actual changes occurring at the database level.**

While the last one states:

The objects in the shared cache that match the delete query **will be invalidated to ensure subsequent persistence contexts do not see the removed objects**.

My read on this is: There is a subtlety as to how delete queries interact with the persistence context in Eclipselink.

praseodym:
In short: when on your branch, execute `git rebase -i origin/master` and a text editor will open. Replace "pick" by "squash" for all commits but the first one, then save and close the text editor. Another text editor will open to let you edit the commit message, make sure to clean it up a little.

More thorough explanation:
  * https://help.github.com/articles/about-git-rebase/
  * https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes

The persistence context is local to the transaction, i.e. the token expiration job in our case. Only if we were to do a batch DELETE somewhere else, we'd have to ensure it is done in an explicit `@Transaction` to ensure consistency with other operations.

jricher:
I'm fine with multi-commit pull requests. FWIW. I'll generally end up doing a rebase before applying to the master branch anyway, so don't sweat it.

strangeweaver:
It's always good to try something new out and I learned about git's view of the world in the process. it's in a single commit now.

strangeweaver:
I've added the changes relating to the 1000 token, limit. 

The behaviour is now that each cleanup operation retrieves a list of items to perform a cleanup operation on and performs the operation on them individually (by default swallowing any exceptions it encounters while performing the operation).

A PageCriteria interface (an interface is used because it leave open the possibility of using say a RandomisedPageCriteria to prevent collisions for example) is used to specify the number of items (page size) to retrieve and the offset to use (page number). Each clean up operation retrieves a page size of 100 and will iterate through a maximum of 1000 pages or run for 10 mins which ever occurs first.

The advantage of this approach is:

- It is resilient to cleanup operations occurring simultaneously in a cluster (#982).
- The max pages/max time/size of page is more configurable I'd like to be able to inject property values but this is subject to the response on (#1022).


Let me know if you see any issues with the latest commits or if you'd rather see them in a separate pull request.

selvarajkumarasamy:
Hello Jricher,
 When you plan to merge this PR to develop,we are facing token cleanup issues,its looking too slow,taking too much time(to delete 1000 token taking 11 minutes),some time getting db error(too many connection).

jricher:
Rebased and applied as 524794fe2e041696b6c3c155c406e8fe097c1c6d, thanks!

I'd like to hear from someone else testing this patch in production to see if it works as advertised. From some initial (small) tests it seems to do the job.

PR #1016, opened 2016-02-08
Allow extenders to add additional information to the access token

dare2try:
While implementing the Session Management spec for MitreID I noticed there was no clean way to provide back with the access token the required claim of `session_state`.  This patch moves the map that was being recreated each time in the `getAdditionalInformation` method up, making it a member of the `OAuth2AccessTokenEntity` so that extenders can can store additional information (like `session_state`) if necessary.


jricher:
This was solved using alternative means, closing the pull request.

PR #1029, opened 2016-02-25
apiAddClient return a HttpStatus.Conflict if you try to create a client with a used client id

trungng92:
This makes `apiAddClient` in the client api return `HttpStatus.Conflict` if you try to create a client with a used client id.

This fixes a bug where if you try to create a client with a client id that is already in use, you get an empty error message. Instead, now you get a message that tells you that the client couldn't be created because the client id is already in use.

To reproduce the bug, you can run the application then:

1. Sign in with `admin`/`password`
2. Go to `Manage Clients` and create a `New Client`
3. Set `Client ID` to a value (in my case I set it to `asdf`)
4. `Save` the client
5. Create a `New Client`
6. Set `Client ID` to the value you put in step 2 (`asdf`)
7. Error message is malformed

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/web/ClientAPI.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/web/ClientAPI.java
@@ -17,10 +17,12 @@
 package org.mitre.openid.connect.web;
 
 import java.lang.reflect.Type;
+import java.sql.SQLIntegrityConstraintViolationException;
 import java.text.ParseException;
 import java.util.Collection;
 import javax.persistence.PersistenceException;
 
+import org.eclipse.persistence.exceptions.DatabaseException;
 import org.mitre.oauth2.model.ClientDetailsEntity;
 import org.mitre.oauth2.model.ClientDetailsEntity.AuthMethod;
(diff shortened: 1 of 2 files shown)


openid-connect-server/src/main/java/org/mitre/openid/connect/web/ClientAPI.java
```diff
@@ -244,6 +245,11 @@ public String apiAddClient(@RequestBody String jsonString, Model m, Authenticati
 			m.addAttribute(HttpCodeView.CODE, HttpStatus.BAD_REQUEST);
 			m.addAttribute(JsonErrorView.ERROR_MESSAGE, "Unable to save client: " + e.getMessage());
 			return JsonErrorView.VIEWNAME;
+		} catch (PersistenceException e) {
```
leonard84: There could be other causes for a `PersistenceException` other than duplication

trungng92: Nice catch. I added checks to make sure that the cause was from a duplicate client id entry.

jricher:
Rebased, tweaked, and committed in ca6e867df6be263b444ce832f9136a8969e3cfd8, thanks!

PR #1032, opened 2016-03-01
stats loading performance issue

zanitete:
This will increase considerably the performance of the stats loading by executing dedicated queries instead of loading the whole entities ApprovedSite and ClientDetailsEntity. It affects many views including:
- site approval page
- client config page


jricher:
Stats are now handled differently throughout the project.

PR #1041, opened 2016-03-16
Fix form action for approve.jsp for root contexts

leonard84:
The current code produces `action="//approve"` for root contexts which will brake the application.

### DIFF
--- a/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/approve.jsp
+++ b/openid-connect-server-webapp/src/main/webapp/WEB-INF/views/approve.jsp
@@ -36,7 +36,7 @@
 		</h1>
 
 		<form name="confirmationForm"
-			action="<%=request.getContextPath()%>/authorize" method="post">
+			action="${request.getContextPath().endsWith('/') ? request.getContextPath() : request.getContextPath().concat('/') }authorize" method="post">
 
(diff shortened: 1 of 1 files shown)


jricher:
Thanks, this is consistent with behavior in the `DiscoveryEndpoint` class, so we'll incorporate it. Interestingly, there are quite a few implementations of this on a root context (like https://mitreid.org/) that don't break today.

praseodym:
Was about to say the same thing, we're running on a root context as well and haven't experienced any problems.

leonard84:
@praseodym yes, there seem to be some setups where it doesn't break. However, it is still a bug and should be fixed.

@jricher this is one point where Thymeleaf would provide better link support with its `@{url}` [syntax](http://www.thymeleaf.org/doc/articles/standardurlsyntax.html). Is there any timeline/roadmap for 1.3?

leonard84:
@jricher I've found a problem if the request parameter is used

> If the OpenId request object parameter is used it hides the implicit
> request object, which in turn causes the expression to fail.
> To circumvent this we need to access the request object through the
> pageContext.

jricher:
Rebased and committed as c96be134daefd58acb6cef57ba4afcab825b950c, thanks!

PR #1043, opened 2016-03-21
Changed from SimpleGrantedAuthority to GrantedAuthority. Suggested fix to #1040

mobilars:
As suggested in issue #1040 I have submitted a pull request where I've changed from simpleGrantedAuthority to GrantedAuthority in the SimpleGrantedAuthorityStringConverter class. I guess given that the SimpleGrantedAuthorityStringConverter class has this name, one could also change the class name and refering files, but I haven't done this. 

This is my first pull request ever, so I hope it works. But don't feel obliged to accept it, it's a suggestion more than a definitive solution. But it does solve my problem here with the cast exception.


ealogar:
Are you planning in merging this PR ?

palmerabollo:
I'm interested, too

PR #1046, opened 2016-03-21
Fix NPE when checking claim extension, Simplify always true expressions

leonard84:
The current implementation does not check for null and will throw a NPE on the `toString()` call.
Furthermore I've fixed some wrong javadoc and removed always true expressions in if clauses.

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/request/ConnectOAuth2RequestFactory.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/request/ConnectOAuth2RequestFactory.java
@@ -17,7 +17,8 @@
 package org.mitre.openid.connect.request;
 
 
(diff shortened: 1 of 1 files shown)


leonard84:
@jricher is there anything blocking this fix?

jricher:
Limited time and ability to work on pull requests.

leonard84:
Is there any timeline when this can be merged and when the next version will be released? Currently we are forced on include a fake `claims` parameter outside of the `request` parameter to avoid this NPE.

jricher:
Rebased and committed as 8e71107f9ba5a43e573388a0ae199079880a50ce, thanks!

leonard84:
@jricher when is the next release planned that contains this fix?

jricher:
We don't have a solid release date in mind at this time. There are a couple additional changes that will likely change the data storage model that we want to get in before locking down 1.3.0, but time is unfortunately limited in pushing this out.

Issue #1056, opened 2016-04-07
DBCS was shown as '?' in pages


liouxiao:
I add Chinese `messages.json` to my own OIDC server, however the JSP pages are not rendered properly. It looks like this:
 
<img width="966" alt="inproper_display" src="https://cloud.githubusercontent.com/assets/3435699/14338670/5a0925d8-fcab-11e5-8ae0-1cdd6b0929eb.png">

liouxiao:
I found the reason is all JSP pages are not set with pageEncoding, so I add `<%@ page language="java" contentType="text/html; charset=utf-8" pageEncoding="utf-8"%>` at the top.

PR #1057, opened 2016-04-07
fix DBCS render issue #1056

liouxiao:
now the page looks like:
proper_display_for_dbcs


jricher:
We're happy to take in this change set, apart from the automated locale resolver portion (see comments on #1058). If you can remove that portion into a separate pull request that would be preferable, otherwise we can back off that change and pull in this request. Thanks!

jricher:

PR #1060, opened 2016-04-07
add Chinese (both simplified and tranditional) messages.

liouxiao:
add zh, zh_CN and zh_TW resources, so that the UI could show Chinese translations.


jricher:

Issue #1061, opened 2016-04-07
The JsonMessageSource can't distinguish locale's country code


liouxiao:
Currently the JsonMessageSource only resolves locale to language level, for example, both `zh_CN` and `zh_TW` are resolved to `zh` (the same issue should be existed when handling `en_US` or `en_GB`). 
However, `zh_CN` and `zh_TW` should use different message resources (for China Mainland, it should be Simplified Chinese; while for Taiwan or HongKong, it should be Traditional Chinese).

PR #1062, opened 2016-04-07
fix issue #1061: auto-detect locale country code

liouxiao:
allow select message resource bundle by full locale (language + country code).


jricher:

PR #1063, opened 2016-04-18
Set encoding utf-8 for application/json response.

fjollberg:
We really should specify an encoding here and not depend on the
servers default encoding, shouldn't we? It becomes ISO-8859-1
otherwise in Tomcat as per the Servlet specification.


jricher:
Rebased and committed as 8f81278332acea7c32554096bc22a90aa25c18bb, thanks!

PR #1067, opened 2016-05-02
Feature/iam 9 convert upstream to spring boot

antoniosignore:


jricher:
Very cool! I'm a little concerned of the duplication of code between this and the -webapp module, is there any way we can pull these in? 

I think we'll target this for development in 1.3.

praseodym:
How do we implement the principle of overlays with Spring Boot? Or will we simply keep building to WAR files?

openid-connect-server-spring-boot/src/main/java/org/mitre/TaskConfig.java
```diff
@@ -0,0 +1,46 @@
...
+    private ApprovedSiteService approvedSiteService;
+	
+	@Bean(destroyMethod="shutdown")
+	public Executor taskScheduler() {
+		return Executors.newScheduledThreadPool(5); // TODO make configureable
+	}
+	
+	// TODO what's exaclty the difference between, taskScheduler and taskExecutor?
+	
+	@Scheduled(fixedDelay = 30000, initialDelay = 60000)
```
leonard84: Shouldn't this use something like `fixedDelayString = "${openid.connect.server.scheduler.token.fixedDelay}"` same for initial

openid-connect-server-spring-boot/src/main/java/org/mitre/CryptoConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.io.Resource;
+
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWEAlgorithm;
+
+@Configuration
+public class CryptoConfig {
+	@Bean(name = "defaultKeyStore")
```
leonard84: Maybe use `@ConditionalOnMissingBean(JWKSetKeyStore.class)` same for all default services, so users can override them if they want to

rfelgent:
Hello folks,

What is the current status of the spring boot migration? Still planning to merge this in the 1.3 branch/master branch ?

PR #1079, opened 2016-06-08
Fix additional info management in access token

marcocaberletti:
Hi,
I fix additionalInfo management in Access Token Entity.
Basically, in the old code, additionalInfo was cleared every time and then the string with the IDToken was added, if it is not empty.
Working on Token Exchange flow, I see that this behavior is a bug: in particular, when I add issued_token_type, the old code lose this required field.
So, I propose this simple fix, that resolves the problem

### DIFF
--- a/openid-connect-client/pom.xml
+++ b/openid-connect-client/pom.xml
@@ -21,7 +21,7 @@
 	<parent>
 		<artifactId>openid-connect-parent</artifactId>
 		<groupId>org.mitre</groupId>
-		<version>1.2.7-SNAPSHOT</version>
+		<version>1.2.7.cnaf-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 	<artifactId>openid-connect-client</artifactId>
--- a/openid-connect-client/pom.xml
+++ b/openid-connect-client/pom.xml
@@ -21,7 +21,7 @@
 	<parent>
 		<artifactId>openid-connect-parent</artifactId>
 		<groupId>org.mitre</groupId>
-		<version>1.2.7.cnaf-SNAPSHOT</version>
+		<version>1.2.7-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 	<artifactId>openid-connect-client</artifactId>
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -21,7 +21,7 @@
 	<parent>
 		<artifactId>openid-connect-parent</artifactId>
 		<groupId>org.mitre</groupId>
-		<version>1.2.7-SNAPSHOT</version>
+		<version>1.2.7.cnaf-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 	<artifactId>openid-connect-common</artifactId>
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -21,7 +21,7 @@
 	<parent>
 		<artifactId>openid-connect-parent</artifactId>
 		<groupId>org.mitre</groupId>
-		<version>1.2.7.cnaf-SNAPSHOT</version>
+		<version>1.2.7-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
 	<artifactId>openid-connect-common</artifactId>
--- a/openid-connect-common/src/main/java/org/mitre/oauth2/model/OAuth2AccessTokenEntity.java
+++ b/openid-connect-common/src/main/java/org/mitre/oauth2/model/OAuth2AccessTokenEntity.java
@@ -64,330 +64,278 @@ import com.nimbusds.jwt.JWT;
 @Entity
 @Table(name = "access_token")
 @NamedQueries({
-  @NamedQuery(name = OAuth2AccessTokenEntity.QUERY_ALL,
-    query = "select a from OAuth2AccessTokenEntity a"),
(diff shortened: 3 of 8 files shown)


jricher:
This field isn't persisted, though, so why would we need to keep this around?

andreaceccanti:
If you are referring to the addtionalInfo field, the reason to "keep it around" is that additionalInfo, AFAIU, is the mechanism used by spring-security-oauth to let token granters decorate the OAuth response. That field is also used by your own code to add the id_token in the response.

andreaceccanti:
Hello,
did the above answer clarify the intention behind this PR? 

Thanks

jricher:
Yes, the field is used to add information to the token response, but why not just re-generate it every time? What's the purpose of saving the generated value if it can be easily re-generated?

andreaceccanti:
Regenerated from what?
The field is not persisted, it's just there to tell spring security oauth
which information should be added to the token response. To implement token
exchange we need to add extra fields to the token response, and this is the
minimum change required to let the token entity class support token
response decoration.

Il giovedì 7 luglio 2016, Justin Richer <notifications@github.com> ha
scritto:

jricher:
Regenerated from whatever source content generated it in the first place. In other parts of the code we use the `extensions` map of the `OAuth2Request` object.

andreaceccanti:
I am not sure I understand your comment, and I actually think that you're
confusing the extension in the request object, which AFAIU is used to handle
additional parameters in the request, with the way spring-security OAuth allows
to decorate the **response**, but I will try to explain more clearly the line
of reasoning behind the proposed change.

Our objective is to implement part of the token exchange spec. To support this
spec, we define a new token granter. The idea is to do some checks on the
request, validate some internal policy that defines which client is allowed to
exchange tokens with which other client, and produce a response in line with
the spec content.

To be compliant we need to add fields to the token response. AFAWK the easiest
way to do this is to use the additionalInformation map in the OAuth2AccessToken
interface:

https://github.com/spring-projects/spring-security-oauth/blob/2.0.10.RELEASE/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/OAuth2AccessToken.java#L70

This is the same approach used by mitreid to add the id token to the "standard"
OAuth response:

https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-common/src/main/java/org/mitre/oauth2/model/OAuth2AccessTokenEntity.java#L147

The above implementation, however, recreates the map at each call to the
getAdditionalInformation method and populates with the id token:

https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-common/src/main/java/org/mitre/oauth2/model/OAuth2AccessTokenEntity.java#L150

This means that the above OAuth2AccessToken implementation cannot be used by
any token granter that wants to decorate the token response, and in particular
cannot be used by our token exchange token granter.

So, given the field is **transient** and, AFAWK, is the only way we have to
decorate the response without writing our own serializer (see below where 
the method is used by the default jackson2 serializer):

https://github.com/spring-projects/spring-security-oauth/blob/2.0.10.RELEASE/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/OAuth2AccessTokenJackson2Serializer.java#L66

we propose to remove the hack that recreates the map at each call to the
getAdditionalInformation method, breaking the contract defined by the
OAuth2AccessToken interface, and instead create an empty map when the
OAuth2AccessTokenEntity is created.

andreaceccanti:
If the above is too long, I include here a TLDR version:

> This field isn't persisted, though, so why would we need to keep this around?

To avoid breaking the contract defined by the OAuth2AccessToken interface.

> Yes, the field is used to add information to the token response, but why not
> just re-generate it every time? What's the purpose of saving the generated
> value if it can be easily re-generated?

There's no saving involved in any part of the code. The field is **transient**.
The map object is generated **anyway** for each token request, as the default
serializer always calls the getAdditionalInformation method. So no memory is saved
with your current approach, but you have the unfortunate side effect that you rule
out the ability for token granters to decorate the token response.

We could have our own OAuth2AccessToken entity implmentation, but I fail to see the
reason given that  one-line change is sufficient to fix OAuth2AccessTokenEntity.

andreaceccanti:
Could this be merged in 1.3.0, or there are objections?

Thanks,

andreaceccanti:
Hello, should we lose all hope on this one?

jricher:
We haven't had time and opportunity to review this pull request in full yet. That said, a similar need was expressed by another user of this project, so we'll eventually come up with something to address it. Right now we're leaning on tying this with changes to how the token responses themselves are being managed. The "additionalInformation" map is a hack to work with the underlying library, and removal of that library is probably a better approach long term. 

So we could add in this pull request temporarily but it would likely get pulled out before 1.3.0 is released.

andreaceccanti:
>  The "additionalInformation" map is a hack to work with the underlying library, and removal of that library is probably a better approach long term.

The additionalInformation is defined by spring-security-oauth as the mechanism to add fields to the token response. The hack currently is the mitre code, and the one-line fix we propose fixes it.

> So we could add in this pull request temporarily but it would likely get pulled out before 1.3.0 is released.

So the plan is for 1.3.0 to drop the spring-security-oauth dependency?

Thanks,

jricher:
Yes, the hack is defined by spring-security-oauth because it conflates the token object with the token response. We've had an issue for removing that library for a while now but haven't been able to get the last few hooks out. We'd like to remove or lessen the dependency in 1.3 if possible, and this is one key location.

jricher:
Concept implemented in 91da3935f575565140f0b648c9d647a94cd2f5d8

PR #1084, opened 2016-06-13
Fix version range for bcprov-jdk15on dependency

gaborcs:
The version range for org.bouncycastle:bcprov-jdk15on was specified incorrectly.
This caused my gradle build to fail when using org.mitre:openid-connect-client:1.2.6 as a dependency.

See version range specification here: http://maven.apache.org/enforcer/enforcer-rules/versionRanges.html


jricher:
Comitted in e08b7f89fcdf42f55f0ef6e403a148828ed5ba0f, thanks!

PR #1092, opened 2016-07-05
shortTopbarTitle

jimmytheneutrino:
Currently, if topbar title is very long, it does not fit nicely on phones. This PR enables one to define a shortTopbarTitle in config, which will replace topbarTitle on phones. This shortTopbarTitle defaults to topbarTitle if undefined, so the change is backward-compatible.


jricher:
Rebased and committed as fa4722cdd8a2d2d3ac770174e43f795b4be443fd, thanks!

PR #1096, opened 2016-07-15
DefaultOAuth2AuthorizationCodeService.createAuthorizationCode should be @Transactional

HeXetic:
An Authentication should not exist without its matching AuthorizationCode, but typically an AuthorizationCode will have a foreign key on an Authentication, meaning it can't be saved first. This block should be wrapped in a transaction so that other DB clients don't see an inconsistent snapshot and then misbehave.

In particular, the cleanup job in DefaultOAuth2ProviderTokenService.clearExpiredTokens can cause data corruption. This may be what was experienced in #982.


jricher:
Thanks, rebased and applied as 7725fcfa2b2ea3625b7df84e6bd8e5cb22375201

PR #1103, opened 2016-07-22
rp jdk6 compatibility

boliang-micropole:
Enabling jdk compilation compatibility in the client component.
Remove java 7 language feature  (diamond generic and global catch)
Upgrade JOSE to 4.22 with a jdk16 classifier
Scoping spring-test dependency to test
It would be nice if the modifications below were made in a maven profile

Scoping the following dependencies to provided in the client component:
- org.apache.httpcomponents:httpclient
- com.google.code.gson:gson
- com.nimbusds:nimbus-jose-jwt
- org.springframework.security.oauth:spring-security-oauth2
- com.google.guava:guava

### DIFF
--- a/openid-connect-client/pom.xml
+++ b/openid-connect-client/pom.xml
@@ -24,6 +24,9 @@
 		<version>1.2.7-SNAPSHOT</version>
 		<relativePath>..</relativePath>
 	</parent>
+	<properties>
+		<maven.javadoc.skip>true</maven.javadoc.skip>
+	</properties>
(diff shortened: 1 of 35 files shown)


jricher:
Java 6 was deprecated several years ago, and we don't support it in recent versions of this project. The base language for 1.2 is Java 7, and the base language for 1.3 is Java 8. We don't have plans to provide backwards compatible support like this.

boliang-micropole:
Ok, I updated the wiki accordingly.

PR #1122, opened 2016-09-16
Configurable HttpClient in DynamicServerConfigurationService

### LINKED ISSUE #1071: Client: Make the used HttpClient configurable to have control over TLS settings
First off, this project is great and has enough flexibility to create your own OP and I really like it. 
Yet, at the moment you cannot configure the HttpClient for the client yourself, and the same code creating the HttpClient, and the RestTemplate are replicated across the client implementation. 
Since you can configure the system properties only once this means that you can only use exactly one key- / truststore per JVM and thus cannot have different PKIs for different OPs. This will also tempt many users to pass in keystone passwords via -D which makes them visible via process info for possible attackers on some OS. 

I would like to propose the introduction of a single method interface RestTemplateFactory#createRestTemplate(String issuerUri) that has a default implementation that matches the current behavior. 
I already made most of the changes but it requires some mildly backwards incompatible changes to the JWKs caches since they only know about the JWK URI and not the associated issuer. This change also de-duplicates some of the code needed to force basic auth / bearer auth and all the places that create a RestTemplate. 

If you're ok with this change I am willing to create a PR. I already have a waiver from my boss but would need to run this change internally though export-control and IP-checks.

jschlichtholz:
Closes #1071

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
+++ b/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
@@ -73,10 +73,15 @@ public class IntrospectingTokenService implements ResourceServerTokenServices {
 	private boolean cacheNonExpiringTokens = false;
 	private boolean cacheTokens = true;
 
-	private HttpClient httpClient = HttpClientBuilder.create()
-			.useSystemProperties()
-			.build();
-	private HttpComponentsClientHttpRequestFactory factory = new HttpComponentsClientHttpRequestFactory(httpClient);
+	private HttpComponentsClientHttpRequestFactory factory;
+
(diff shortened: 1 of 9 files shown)


jschlichtholz:
This just fixes #1071 for the DynamicServerConfigurationService.

@jricher if you agree with the general approach I would update the PR and add a similar change to the other usages of HttpClient.

jricher:
I haven't tested this code but it looks reasonable to me on first view. Feel free to expand the pull request to cover other areas of the code as well. Thank you!

jschlichtholz:
Done. The only change I made, aside from what I suggested previously, was to enable injection of a HttpClient into OIDCAuthenticationFilter.

jricher:
Rebased and applied as c3d0c18af5c5d3c4fe226d2ef998d71e64d873c1, thanks!

PR #1125, opened 2016-09-21
use the same encoding as on client side

ngriesser:
see OIDCAuthenticationFilter line 336


jricher:
Thanks, also applied to 1.3 as f45a6ef56a6a51e28272e24959ab26da7caeeef2

PR #1126, opened 2016-09-24
Fixed NPE in case if algorithm is not specified

Shpota:
In case if token endpoint auth method is set to client_secret_jwt and signing algorithm is not specified the system fails with NPE in  `org.mitre.openid.connect.client.OIDCAuthenticationFilter#handleAuthorizationCodeResponse`

There is a predefined validation that should work in this case. The validation checks signer service and throws AuthenticationServiceException with human readable message "Couldn't find required signer service for use with private key auth". 
The issue is that the system fails with NPE before entering into the validation code.


jricher:
Rebased and applied as 476ec872ff0f2caddb684c80b5802af5ccab7883, thanks!

Reading through the code it seems like all branches are still covered appropriately by the other checks.

PR #1137, opened 2016-10-25
Fix DefaultIntrospectionResultAssembler and IntrospectingTokenService to populate and parse `user_id` according to user authentication

sofiaguyang:
The problem is in both `DefaultIntrospectionResultAssembler` and `IntrospectingTokenService`

With `DefaultIntrospectionResultAssembler`, it populates `user_id` with OAuth2Authentication#getName() which in turn uses OAuth2Authentication#getPrincipal() that defaults to the client id if user authentication is not available. This leads to the `user_id` being falsely populated with the client_id when user authentication is not available. 

With `IntrospectingTokenService`, it uses `sub` which cannot be used to create the user authentication because it may not necessarily refer to the user. Instead if may refer to the client
if the access token happens to be client-only.

This merge request addresses these problems.

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
+++ b/openid-connect-client/src/main/java/org/mitre/oauth2/introspectingfilter/IntrospectingTokenService.java
@@ -235,8 +235,13 @@ public class IntrospectingTokenService implements ResourceServerTokenServices {
 		return storedRequest;
 	}
 
-	private Authentication createAuthentication(JsonObject token) {
-		return new PreAuthenticatedAuthenticationToken(token.get("sub").getAsString(), token, introspectionAuthorityGranter.getAuthorities(token));
+	private Authentication createUserAuthentication(JsonObject token) {
+		JsonElement userId = token.get("user_id");
+		if(userId == null) {
+			return null;
+		}
+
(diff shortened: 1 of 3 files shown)


jricher:
In a client credentials flow, the subject of the token is in fact the client itself, so that part is intentional.

sofiaguyang:
@jricher Yes that's right. So, in the client credentials flow, the sub here will still display the client id. However, the user_id displays also shows the client_id. Here's an example of an introspection result I have:

```
{
  "active": true,
  "scope": "internal hey1 openid hey2 offline_access bill",
  "expires_at": "2016-10-25T09:34:03+0800",
  "exp": 1477359243,
  "sub": "6e1a704b-0eba-4ff0-89d5-ade9762e2705",
  "user_id": "6e1a704b-0eba-4ff0-89d5-ade9762e2705",
  "client_id": "6e1a704b-0eba-4ff0-89d5-ade9762e2705",
  "token_type": "Bearer"
}
```

The user_id here should be null and absent.

jricher:
I think that's a fair interpretation and I'm not against pulling this one in to the dev branch. You'll need to fix the unit tests first, though.

sofiaguyang:
@jricher Yup, will fix the test. Thanks!

sofiaguyang:
@jricher I've fixed the tests. Thanks!

jricher:
Rebased and applied as bb6bb81dbcfc8c1d835c437bedaa70e51f2e966f, thanks!

PR #1143, opened 2016-11-04
Oracle support added

tborowiec:


openid-connect-server-webapp/src/main/resources/db/clients_oracle.sql
```diff
@@ -0,0 +1,51 @@
+--
```
tborowiec: New file (not using `clients.sql`) because:
* Oracle doesn't support mass INSERTs
* Oracle doesn't have a BOOLEAN type
* Oracle's MERGE syntax is slightly different

openid-connect-server-webapp/src/main/resources/db/tables/loading_temp_tables_oracle.sql
```diff
@@ -0,0 +1,77 @@
+--
```
tborowiec: New file (not using `loading_temp_tables.sql`) because:
* same reasons as for `clients.sql`
* additionally, Oracle has a different syntax for creating temporary tables

openid-connect-server-webapp/src/main/resources/db/tables/oracle_database_tables.sql
```diff
@@ -0,0 +1,402 @@
+--
```
tborowiec: Changes comparing to `hsql_database_tables.sql`:
* Oracle doesn't have a BIGINT type - changed it to NUMBER(19)
* Oracle doesn't support identity - changed it to separate sequence for every numeric ID
* Oracle doesn't have a BOOLEAN type - changed it to NUMBER(1) and added constraint check for 0 or 1 for every boolean
* some identifiers exceed Oracle's 30 character limit, so they needed to be shortened (please see `entity-mappings_oracle.xml`)
* Oracle's VARCHAR2 type doesn't support columns larger than 4000 characters, so for 4096 character columns I changed it to 4000, for 8192 I changed the datatype to CLOB

openid-connect-server-webapp/src/main/resources/entity-mappings_oracle.xml
```diff
@@ -0,0 +1,266 @@
+<entity-mappings xmlns="http://xmlns.jcp.org/xml/ns/persistence/orm"
```
tborowiec: I used this file to change values provided in java annotations for entities - that way we do not need to modify the code to get it working for oracle. Two types of changes here:
* sequence generator insteqad of identity for IDs
* identifiers exceeding 30 characters shortened

jricher:
Rebased and added in 83a9fef14dd4d29ca456c87f3ed2147a96e71e30 (et. al). Thanks!

Issue #1133, opened 2016-10-14
Fix field definition in the model object


JessicaCar:
Hello, 

We are developing our SSO with Mitreid OpenID Connect. We have detected that ACR Default Values field is not saved in database from the Edit client screen . We have seen that it is because in openid-connect-server-webapp / src / main / webapp / resources / js / Client.js the field  defaultAcrValuesCollection and defaultAcrValues should be called defaultACRvaluesCollection and defaultACRvalues to be saved correctly in database.

You want we do push for change or want to do it you.

Thank you very much.

Jessica Carpintero.

jricher:
This would be great as a pull request, or we can try to fix it. This wasn't caught as nothing in the core server uses that field yet.

PR #1161, opened 2016-12-13
Fix field definition in the client model object.

juazugas:
Hi! 

This pull request is for the issue 'Fix field definition in the model object #1133', Jessica wrote: 

> 'We have detected that ACR Default Values field is not saved in database from the Edit client screen . We have seen that it is because in openid-connect-server-webapp / src / main / webapp / resources / js / Client.js the field defaultAcrValuesCollection and defaultAcrValues should be called defaultACRvaluesCollection and defaultACRvalues to be saved correctly in database.'

We have corrected both variables (defaultACRvaluesCollection and defaultACRvalues) to keep the relation with the model. Now the json maps to the client correct value.


jricher:
Rebased and applied as ffc5782810f62d0a87db8c96ae thanks!

PR #1174, opened 2017-01-12
Fixed overlay problems in uma-server-webapp

sschu:
I was trying to run the UMA server and experienced some problems coming from overlayed files no longer working correctly with base files. I adapted the overlayed files according to the base files which fixed the problems for me. However, I have not tested it thoroughly so I am not sure everything is working 100%, especially the security configuration part...


jricher:
Rebased and committed in 0c09a17f59f59372a7263df86b0ab995a9cbc261 (et al) thanks!

PR #1187, opened 2017-02-03
Change Base64URL encoding for pkce checking

shingz96:
Use Java 1.8 Base64.getUrlEncoder() to encode instead of com.nimbusds.jose.util.Base64URL.
As com.nimbusds.jose.util.Base64URL will remove the "=" at the end while Base64URL encoding only needed to remove trailling "="


jricher:
What change is actually being made here? They both remove the trailing `=` sign, which is required.

shingz96:
But like `=abc=` will be encode into `abc ` (with com.nimbusds.jose.util.Base64URL).
Should be encode as `abc=` in Base64URL encoding.

jricher:
No, all of the padding `=` characters should be stripped from the tail of the string for this purpose. The current code is correct, in that case.

PR #1192, opened 2017-02-17
Add parameters from error response to exception

hdpe:
It would be useful to throw a richer exception due to an authentication error response in the client, encapsulating the parameters from the response, rather than just throwing an `AuthenticationServiceException` with a concatenated string of the parameters as its message.

This prevents having to re-retrieve the parameters in the `AuthenticationFailureHandler`, or allows you to use a redirect in a `SimpleUrlAuthenticationFailureHandler` and still display the error response parameters in the final page by retrieving the `org.springframework.security.web.WebAttributes.AUTHENTICATION_EXCEPTION` session attribute. 

The default behaviour of `SimpleUrlAuthenticationFailureHandler` is to do a redirect, which means the query parameters would have been lost. This failure handler can be configured to use a forward instead via `setUseForward()` but a redirect may be preferable for various reasons.

### DIFF
--- a/openid-connect-client/src/main/java/org/mitre/openid/connect/client/AuthorizationEndpointException.java
+++ b/openid-connect-client/src/main/java/org/mitre/openid/connect/client/AuthorizationEndpointException.java
@@ -1,0 +1,33 @@
+package org.mitre.openid.connect.client;
+
+import org.springframework.security.authentication.AuthenticationServiceException;
+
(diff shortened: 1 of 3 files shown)


jricher:
Rebased and committed as e1ae8f3d8da6efa0e1c03632d9c8e9117a42dfcd, Thanks!

PR #1199, opened 2017-03-02
Fix  NPE if no claims are requested for the userinfo object

leonard84:
This happens if clients only requests id_token claims, or just send an empty claims parameter.

Change-Id: I8bd176ad271bda8a1e2f26b6221bd8e2d0a3ebfb


jricher:
Cleaned up a little bit and committed as 00ecd3dd22a1863a2607b748546547666315252f and 98a4d56cddcd8b9a93d4cf640bc3a1c8a07633b7

Thanks!

PR #1204, opened 2017-03-14
fix XML validation a bit

thomasmey:
Only remaining problems:
```
            <element-collection fetch="EAGER" name="authorities">
                <column name="authority"/>
                <convert converter="org.mitre.oauth2.model.convert.SimpleGrantedAuthorityStringConverter"/>
                <collection-table name="auth_holder_authority">
                    <join-column name="owner_id"/>
                </collection-table>
            </element-collection>

```
according to XSD it's either column or converter in an element-collection, so what's the correct way of defining this in the entity-mapping file?
See http://www.utilities-online.info/xsdvalidation/ output

### DIFF
--- a/openid-connect-server-webapp/src/main/resources/db/oracle/entity-mappings_oracle.xml
+++ b/openid-connect-server-webapp/src/main/resources/db/oracle/entity-mappings_oracle.xml
@@ -263,4 +263,26 @@
         </attributes>
     </entity>
 
+    <entity class="org.mitre.openid.connect.model.DefaultAddress" name="DefaultAddress">
+        <attributes>
+            <!-- changing generated value to sequence strategy (Oracle doesn't support identity) -->
+            <id name="id">
+                <column name="id"/>
+                <generated-value strategy="SEQUENCE" generator="DefaultAddressSequenceGenerator"/>
+                <sequence-generator name="DefaultAddressSequenceGenerator" sequence-name="address_seq" allocation-size="1"/>
+            </id>
+        </attributes>
+    </entity>
+
+    <entity class="org.mitre.openid.connect.model.DefaultUserInfo" name="DefaultUserInfo">
+        <attributes>
+            <!-- changing generated value to sequence strategy (Oracle doesn't support identity) -->
+            <id name="id">
+                <column name="id"/>
+                <generated-value strategy="SEQUENCE" generator="DefaultUserInfoSequenceGenerator"/>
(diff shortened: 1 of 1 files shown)


jricher:
It doesn't make sense that you wouldn't have both column and converter on a class, as the annotation version of this is:

```java
	/**
	 * @return the authorities
	 */
	@ElementCollection(fetch = FetchType.EAGER)
	@CollectionTable(
			name="authentication_holder_authority",
			joinColumns=@JoinColumn(name="owner_id")
			)
	@Convert(converter = SimpleGrantedAuthorityStringConverter.class)
	@Column(name="authority")
	public Collection<GrantedAuthority> getAuthorities() {
		return authorities;
	}
```

As for the rest of the changes, they mostly seem to be about tag order and shouldn't matter, though I'm happy to take in the pull request if people want. I don't have an Oracle database to test against so I can't verify it myself.

thomasmey:
Yes, this PR just reorders the xml elements, so it validates against the XSD schema. My editor did show warnings for the element order in the xml file, as the XSD ordering is a strict one.

Regarding @Converter and @Column: Strangely the XSD defines it this way:

      &lt;xsd:choice&gt;
        &lt;xsd:sequence&gt;
          &lt;xsd:element name="column" type="orm:column" minOccurs="0"/&gt;
          &lt;xsd:choice&gt;
            &lt;xsd:element name="temporal"
                         type="orm:temporal"
                         minOccurs="0"/&gt;
            &lt;xsd:element name="enumerated"
                         type="orm:enumerated"
                         minOccurs="0"/&gt;
            &lt;xsd:element name="lob"
                         type="orm:lob"
                         minOccurs="0"/&gt;
          &lt;/xsd:choice&gt;
        &lt;/xsd:sequence&gt;
        &lt;xsd:sequence&gt;
          &lt;xsd:element name="attribute-override"
                       type="orm:attribute-override"
                       minOccurs="0" maxOccurs="unbounded"/&gt;
          &lt;xsd:element name="association-override"
                       type="orm:association-override"
                       minOccurs="0" maxOccurs="unbounded"/&gt;
          &lt;xsd:element name="convert" type="orm:convert"
                       minOccurs="0" maxOccurs="unbounded"/&gt;
        &lt;/xsd:sequence&gt;
      &lt;/xsd:choice&gt;

so it's either an column element or an convert element, not both. maybe you need to define two element-collection elements for the same name and one must contain the column and one must contain the converter.

PR #1205, opened 2017-03-18
DefaultUserInfoService: Optimize access path

thomasmey:
When no user_info is availble skip, the db access for the client
details.


jricher:
This interceptor is only used for various parts of the UI, so in fact it could be skipped for any of the API or backend endpoint URLs.

jricher:
Addressed more thoroughly by 52829d4adbd90b2dc624be1fd0a0bd58d20bf3da, thanks!

PR #1206, opened 2017-03-19
UserInfoInterceptor: Exclude interceptor for /introspect endpoint

thomasmey:
The user_info is never used in the introspection endpoint. Avoid a
roundtrip to the database here.


jricher:
The `UserInfo` is used in this call (the user's subject is injected into the results), but it's not the object from the interceptor. The introspection endpoint does its own lookup in the results assembler, if necessary.

jricher:
Implemented with a wider net in 52829d4adbd90b2dc624be1fd0a0bd58d20bf3da, thanks!

thomasmey:
Thanks for merging :-)

PR #1207, opened 2017-03-19
Statically weave some entity classes

thomasmey:
This dramatically reduces the number of SQLs that are issued by the JPA
framework for the DefaultIntrospectionAssembler case in the
IntrospectionEndpoint.

### DIFF
--- a/openid-connect-client/pom.xml
+++ b/openid-connect-client/pom.xml
@@ -32,6 +32,11 @@
 			<groupId>org.mitre</groupId>
 			<artifactId>openid-connect-common</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.persistence</groupId>
+			<artifactId>org.eclipse.persistence.jpa</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 	<packaging>jar</packaging>
 	<build>
--- a/openid-connect-common/pom.xml
+++ b/openid-connect-common/pom.xml
@@ -86,6 +86,11 @@
 			<groupId>org.bouncycastle</groupId>
 			<artifactId>bcprov-jdk15on</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.persistence</groupId>
+			<artifactId>org.eclipse.persistence.jpa</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<packaging>jar</packaging>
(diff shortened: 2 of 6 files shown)


jricher:
This looks like it's mostly turning on lazy loading for class elements, is there something else going on that's really making a difference?

thomasmey:
Yes, the change statically weaves the eclipse link classes and enables lazy attribute access.
This reduces the number of SQLs for the /introspection endpoint from 30-something to 2-4 depending on the code path.
The greatly reduces the response time for this endpoint, when heavily used.

jricher:
It would make sense to help in that one instance since the introspection endpoint doesn't make use of any of those class attributes to do its job (in the normal case) apart from Scope. Are there any effects on other parts of the system which *do* use those attributes?

I noticed that you've also submitted changes to the Oracle setup, so has this been tested on other databases as well (HSQL, MySQL, Postgres)?

thomasmey:
"Are there any effects on other parts of the system which do use those attributes?"
Theoretically no, it only makes the access more dynamic in the sense that the database is only accessed when the data is really needed. Compared to the current state where everything is always read from the database.

"I noticed that you've also submitted changes to the Oracle setup, so has this been tested on other databases as well (HSQL, MySQL, Postgres)?"

I tested this change specifically against Postgres 9.6.1 only.
The weaving should theoretically be independent from the database version used.

jricher:
If I'm understanding this correctly, then, we could get the same results you're after by simply using lazy loading and keeping dynamic weaving. The static weaving shouldn't affect the performance. Besides, it's been turned off in jpa-config.xml at the moment:

https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/8406a89fd12a09929200beadf3ec5459c5b5057a/openid-connect-server-webapp/src/main/webapp/WEB-INF/jpa-config.xml#L44

I'm hesitant to have another configuration file (persistence.xml) listing model classes as that's another element to get out of sync with the rest of the system over time, unless it's really worth it.

thomasmey:
Yes, correct. Besides that as far as I understand dynamic weaving isn't compatible with the byte code manipulation going on in Spring framework.
Spring and dynamic weaving is only possible by using a spring JVM agent. Static weaving has the advantage that it works out of the box in all applications servers without using some special JVM parameters and possible jar libraries outside the war file.

Yes, the extra confit file is just used for the maven plugin to list all entities needing weaving. But at least the plugin stops the maven build with an error of something doesn't match, so errors in the case of refactoring should be easily catched.

PR #1211, opened 2017-03-22
Decrease log level of trailing slash warning

praseodym:
Having an issuer without trailing slash configured is just fine, so
there is no reason to log a warning for this every time the discovery
endpoint is called.


jricher:
Rebased and included as 903168a94917487c6da53103051f5a64115caa0c, thanks!

PR #1221, opened 2017-04-04
Replace cobertura with jacoco to generate code-coverage for codecov.io

leonard84:


jricher:
Rebased, cleaned up, and check in as 2aecedfb3d273552e15d41b702ddc87e1b168025, thanks!

PR #1251, opened 2017-06-07
change BIGINT AUTO_INCREMENT to SERIAL

joegalley:


jricher:
Already covered by #1228

PR #1266, opened 2017-07-10
Only log mismatch warnings, if values are actually defined and differ

leonard84:
Currently MITRE always logs a mismatch warning when a value is only defined in the request object. However, it should only log a warning if a value is defined both in the request object and the authorization request and their values differ.


leonard84:
@jricher what do you think about this? If you are ok with it, I could see about adding some tests.

jricher:
Not all of the fields are optional as per the spec, so some really should have stricter checks. We've tended to be more relaxed in our implementation than strictly necessary:

>  So that the request is a valid OAuth 2.0 Authorization Request, values for the response_type and client_id parameters MUST be included using the OAuth 2.0 request syntax, since they are REQUIRED by OAuth 2.0. The values for these parameters MUST match those in the Request Object, if present.

> Even if a scope parameter is present in the Request Object value, a scope parameter MUST always be passed using the OAuth 2.0 request syntax containing the openid scope value to indicate to the underlying OAuth 2.0 logic that this is an OpenID Connect request.

If those can be fixed up, it makes sense to include. We'll always welcome more test coverage!

PR #1281, opened 2017-07-24
Missing bootstrap css files in webapp/resources

janhelge:
Added missing bootstrap css files to openid-connect-server-webapp.


jricher:
These files are generated upon build by the LESS compiler and don't need to be included in the static build. They were removed in c377e411e390ad13ea4fd081d538751a583b808b

PR #1291, opened 2017-08-16
Fix mysql tables

LorenzReinhart:
Added temp_tables in mysql. Added missing values in tables. Fixed insertion of duplicate user info.


silentsudo:
I am getting some errors i think this pr has its solution

exodussystem:
This caused issue with the JPA mapping that maps client's scopes, redirect URIs and grant types using sequence ID (numeric value).

PR #1320, opened 2017-11-13
Client: Parse 'sub' key to identify resource owner in introspection response (RFC7662)

bverhoeven:
First of all: Thanks for your work! Getting the OpenID Connect client library to work was surprisingly easy. 

As per https://tools.ietf.org/html/rfc7662#section-2.2 the `sub` key should identify the resource owner in oauth2 introspection responses:

>    sub
      OPTIONAL.  Subject of the token, as defined in JWT [RFC7519].
      Usually a machine-readable identifier of the resource owner who
      authorized this token.

I've found that this OpenID Connect client library checks for the (non-standard?) `user_id` key, which prevents responses from RFC-compliant servers to be parsed. 

This pull request adds support for the `sub` key and will allow the introspection response of RFC-compliant servers to be parsed. Will still try `user_id` first as to not break backward compatibility.


jricher:
That's really interesting, I thought for sure that had been changed in this code when the RFC was published (since I'm the editor for both, this is somewhat embarassing). Thanks for the catch, we'll get this rolled in!

PR #1330, opened 2017-12-04
Add GrantedAuthority converter to avoid errors when client has authorities

ifernandezautentia:
Good afternoon.

The following request resolves an error that occurs when the detail of a client is consulted and it has associated roles.

This is because the json deserialiser is not capable of generating the GrantedAthorities through the json received at the controller.

The error has been solved by associating a converter to the gson builder that will be responsible for transofromating the json to Java classes.


jricher:
Wouldn't it be simpler to serialize these as strings in an array instead of objects with a single member?

PR #1337, opened 2017-12-21
Migration to Spring OAuth 2.2.x

splatch:
Beside a  version bump pull request provides some cleanup of POM elements which were redundant or missplaced improving project descriptor readability.


jricher:
I'm not a fan of how this pulled most of the version numbers out of the plugin management section and into properties on the project. It seems unnecessary given that most of these are single dependencies.

splatch:
That's why I separated commits into two - first which lift up Spring alone, second which does cleanups.

It's up to you if you pick second one. Placeholder for wro4j you can be overriden via `-Dwro4j.version` passed to build. Not sure if anyone will ever use this. If you want to keep wro4j version static in dependency management - I can ammend commit so it stays as is.

jricher:
It's not just WRO4J, it's everything. I'm not sure why anyone would want to override dependency versions on the commandline.

PR #1346, opened 2018-01-20
Fix interface for issuer URI without trailing slash

### LINKED ISSUE #1345: Admin interface fails with issuer URL does not have a trailing slash
When issuer URL is configured as e.g. `http://example.com` (no trailing slash), the admin interface (`admin.js`) will try to request `http://example.com.well-known/openid-configuration` which does not work.

praseodym:
Fixes #1345


praseodym:
Updated to also fix the statistics panel on the home page.

PR #1355, opened 2018-02-07
Throwing exception on all other JWT types than SignedJWT

tborowiec:


openid-connect-server/src/main/java/org/mitre/openid/connect/assertion/JWTBearerAuthenticationProvider.java
```diff
@@ -91,57 +91,60 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+				if (config.isHeartMode() && !client.getTokenEndpointAuthMethod().equals(AuthMethod.PRIVATE_KEY)) {
+					throw new AuthenticationServiceException("[HEART mode] Invalid authentication method");
+				}
 
-				JWSAlgorithm alg = jws.getHeader().getAlgorithm();
+				JWTSigningAndValidationService validator = validators.getValidator(client, alg);
 
-				if (client.getTokenEndpointAuthSigningAlg() != null &&
-						!client.getTokenEndpointAuthSigningAlg().equals(alg)) {
-					throw new InvalidClientException("Client's registered request object signing algorithm (" + client.getRequestObjectSigningAlg() + ") does not match request object's actual algorithm (" + alg.getName() + ")");
```
tborowiec: Exception class and message was incorrect - this was probably a copy-paste from ConnectOAuth2RequestFactory. The problem was that if this was thrown, it was then caught in this method and UsernameNotFoundException was thrown with a message that client couldn't be found.

PR #1365, opened 2018-02-21
Catch Exception

aishgosavi:
Tried to test if the exception is thrown

### DIFF
--- a/openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
+++ b/openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
@@ -149,8 +149,8 @@ public class TestDefaultIntrospectionResultAssembler {
 		assertThat(result, is(equalTo(expected)));
 	}
 
-	@Test
-	public void shouldAssembleExpectedResultForAccessTokenWithoutExpiry() {
+	@Test (expected = ParseException.class)
+	public void shouldAssembleExpectedResultForAccessTokenWithoutExpiry() throws Throwable {
 
 		// given
 		OAuth2AccessTokenEntity accessToken = accessToken(null, scopes("foo", "bar"), null, "Bearer",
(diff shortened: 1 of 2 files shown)


jricher:
You have several syntax errors in your code. Have you compiled and run this locally?

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -149,8 +149,8 @@ public void shouldAssembleExpectedResultForAccessTokenWithoutUserInfo() throws P
 		assertThat(result, is(equalTo(expected)));
 	}
 
-	@Test
-	public void shouldAssembleExpectedResultForAccessTokenWithoutExpiry() {
+	@Test (expected = ParseException.class)
```
jricher: This test should not throw an exception, so why are you saying that an exception was expected? Leave this test as-is.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -174,8 +174,23 @@ public void shouldAssembleExpectedResultForAccessTokenWithoutExpiry() {
...
 				.build();
 		assertThat(result, is(equalTo(expected)));
+		assertFalse(throwException());
+	}
+	
+	private void assertFalse(boolean throwException) {
+		
 	}
 
+	private boolean throwException() throws ParseException{
```
jricher: You shouldn't need to throw your own exceptions, you're trying to see if the underlying code is going to throw an exception in the right circumstances.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -255,8 +270,8 @@ public void shouldAssembleExpectedResultForRefreshTokenWithoutUserInfo() throws
 		assertThat(result, is(equalTo(expected)));
 	}
 
-	@Test
-	public void shouldAssembleExpectedResultForRefreshTokenWithoutExpiry() {
+	@Test(expected = ParseException.class)
+	public void shouldAssembleExpectedResultForRefreshTokenWithoutExpiry() Throwable{
```
jricher: Same comment as above, this test should not throw an exception. To test the code that will throw an exception, create a new `@Test`.

PR #1366, opened 2018-02-22
Created TestDefaultOAuth2AuthorizationCodeService

anand10parmar:
Hi Justin,
Please let me know what will be the correct direction. i have mock different classes to get my test work. Because of dependency i tried to do this.
Also tomorrow i will be working on it whole day.
Hope i will get this soon. also tomorrow we will be having peer meeting to solve doubt.

### DIFF
--- a/openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
+++ b/openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
@@ -1,0 +1,53 @@
+package org.mitre.oauth2.service.impl;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.when;
+import org.junit.Test;
+import org.mitre.oauth2.model.AuthenticationHolderEntity;
+import org.mitre.oauth2.model.AuthorizationCodeEntity;
+import org.mitre.oauth2.repository.AuthorizationCodeRepository;
+import org.mitre.oauth2.repository.impl.JpaAuthenticationHolderRepository;
+import org.mitre.oauth2.repository.impl.JpaAuthorizationCodeRepository;
+import org.mockito.Matchers;
+import org.mockito.Mockito;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
+
+public class TestDefaultOAuth2AuthorizationCodeService {
+
+	@Test
+	public void testConsumeAuthorizationCode()
+	{
+
+DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
+AuthenticationHolderEntity authHolder = Mockito.mock(AuthenticationHolderEntity.class);
+OAuth2Authentication authentication = Mockito.mock(OAuth2Authentication.class);
+JpaAuthenticationHolderRepository authRepo = Mockito.mock(JpaAuthenticationHolderRepository.class);
+
+doNothing().when(authHolder).setAuthentication(Matchers.isA(OAuth2Authentication.class));
+when(authRepo.save(Matchers.isA(AuthenticationHolderEntity.class))).thenReturn(authHolder);
+AuthorizationCodeRepository repository = Mockito.mock(AuthorizationCodeRepository.class);
+AuthorizationCodeEntity entity = Mockito.mock(AuthorizationCodeEntity.class);
+when(repository.save(Matchers.isA(AuthorizationCodeEntity.class))).thenReturn(entity);
+String code = defaultAuth.createAuthorizationCode(authentication);
+assertEquals(repository.getByCode(code),entity);
+
+	}
+	
+	
+	@Test
+	public void testRespositoryGetAndSet() {
+		AuthorizationCodeRepository repository = new JpaAuthorizationCodeRepository() ;
+		DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
+		defaultAuth.setRepository(repository);
+		assertEquals(defaultAuth.getRepository(), repository);			
+	}
+	
+	@Test
+	public void testExpirationSecond() {
+		DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
+		defaultAuth.setAuthCodeExpirationSeconds(10);
+		assertEquals(defaultAuth.getAuthCodeExpirationSeconds(), 10);			
+	}
+	
+}


openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,53 @@
...
+import org.mitre.oauth2.model.AuthenticationHolderEntity;
+import org.mitre.oauth2.model.AuthorizationCodeEntity;
+import org.mitre.oauth2.repository.AuthorizationCodeRepository;
+import org.mitre.oauth2.repository.impl.JpaAuthenticationHolderRepository;
+import org.mitre.oauth2.repository.impl.JpaAuthorizationCodeRepository;
+import org.mockito.Matchers;
+import org.mockito.Mockito;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
+
+public class TestDefaultOAuth2AuthorizationCodeService {
```
jricher: This needs to be annotated with the appropriate test runners to be picked up by the test framework, see the other unit test classes

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,53 @@
...
+
+	@Test
+	public void testConsumeAuthorizationCode()
```
jricher: the formatting of this function is off, please format like other code in the project before submission

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,53 @@
...
+	{
+
+DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
+AuthenticationHolderEntity authHolder = Mockito.mock(AuthenticationHolderEntity.class);
```
jricher: these mocks can be made using `@Mock` annotations and injected automatically into the service instead of being set directly; this is the preferred method as this mimics what Spring does at runtime

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,53 @@
...
+OAuth2Authentication authentication = Mockito.mock(OAuth2Authentication.class);
+JpaAuthenticationHolderRepository authRepo = Mockito.mock(JpaAuthenticationHolderRepository.class);
+
+doNothing().when(authHolder).setAuthentication(Matchers.isA(OAuth2Authentication.class));
+when(authRepo.save(Matchers.isA(AuthenticationHolderEntity.class))).thenReturn(authHolder);
+AuthorizationCodeRepository repository = Mockito.mock(AuthorizationCodeRepository.class);
+AuthorizationCodeEntity entity = Mockito.mock(AuthorizationCodeEntity.class);
+when(repository.save(Matchers.isA(AuthorizationCodeEntity.class))).thenReturn(entity);
```
jricher: This tells the repository mock to return the same entity for every "save" call, which doesn't really test what we want it to test. Instead, we're looking to make sure that the repository's "save" method is called with an authorization code with the right parameters -- specifically, with the associated authentication. This should be true even if we call it with separate authentications, which isn't happening here. 

Look into using JUnit/Mockito's "Answer" construct instead of straight mocks for this.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,53 @@
...
+	public void testRespositoryGetAndSet() {
+		AuthorizationCodeRepository repository = new JpaAuthorizationCodeRepository() ;
+		DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
+		defaultAuth.setRepository(repository);
+		assertEquals(defaultAuth.getRepository(), repository);			
+	}
+	
+	@Test
+	public void testExpirationSecond() {
+		DefaultOAuth2AuthorizationCodeService defaultAuth = new DefaultOAuth2AuthorizationCodeService();
```
jricher: This is helpful as a simple set but doesn't test whether or not the authorization code return has the correct timeout for it, please create another test for that:

  - create one authz code with the default, check its timeout
  - set the timeout to something different and obvious (99 minutes or something?)
  - create a second authz code with a new authentication object and check its timeout

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
jricher: Note this from our code coverage, this says that your code isn't being called by the unit test framework:

image

anand10parmar:
Kindly guide me with this commit.
I tried making changes as per you said.

jricher:
You did some of the things I pointed to but not others, there's still work to do. There's also the biggest function still marked as "TODO"

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,129 @@
...
+	@Mock
+	private AuthorizationCodeRepository repository;
+	
+	@Mock
+	private AuthenticationHolderRepository authenticationHolderRepository;
+	
+	
+	//following two mock suggested by justin
+	@Mock
+	private JpaAuthenticationHolderRepository  authRepo;
```
jricher: you don't need this mock and the one above it -- use only the first of these two (that uses the interface instead of a concrete class)

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,129 @@
...
+		}
+	
+	@Test
+	public void testRespositoryGetAndSet() {
+		service.setRepository(repository);
+		assertEquals(service.getRepository(), repository);			
+	}
+	
+	@Test
+	public void testExpirationSecond() {
```
jricher: Add another test that checks whether the authorization code that is saved is set with the appropriate seconds. So something like this horrible pseudocode: 

service.setAuthCodeExpiration(99)
verify(respository.save(argumentcaptor))
assertThat(argumentcaptor.expiration().isEqualTo(Date.now() + 99seconds)

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	@Test
+	public void testConsumeAuthorizationCodeNotFound() {
+		assertThat(null, equalTo(service.consumeAuthorizationCode(codeNotFound)));	
+	}
+	
+
+	@Test
+	public void testConsumeAuthorizationCode(){
```
jricher: add another test that shows an authorization code can't be consumed twice; try it twice and expect an error on the second one

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,129 @@
...
+	@Test
+	public void testCreateAuthorizationCode(){
+		
+		OAuth2Request auth2Request = Mockito.mock(OAuth2Request.class);
+		Authentication authObj = Mockito.mock(Authentication.class);
+		OAuth2Authentication authentication = new OAuth2Authentication(auth2Request, authObj);
+		String code = service.createAuthorizationCode(authentication);
+		ArgumentCaptor<AuthorizationCodeEntity> argumentCaptor = ArgumentCaptor.forClass(AuthorizationCodeEntity.class);
+		Mockito.verify(repository).save(argumentCaptor.capture());
+		assertThat(argumentCaptor.getValue(), equalTo(code));
```
jricher: this looks fine

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
anand10parmar: Thanks Justin for the pseudo.
Will you please tell me what is .expiration() here.

jricher: It's the expiration of the authorization code that you're testing, you need to make sure that it expires off in the future by the correct amount.

anand10parmar: Thanks, Justin.
I have made a new test as per your suggestion for expiration code. kindly go through it my recent commit.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.security.oauth2.common.exceptions.InvalidGrantException;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
+import org.springframework.security.oauth2.provider.OAuth2Request;
+
+
+
+@RunWith(MockitoJUnitRunner.class)
+public class TestDefaultOAuth2AuthorizationCodeService {
+	
+	public int counter = 0;
```
jricher: What is this for and why is it public?

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,135 @@
...
+	
+	@Test
+	public void testCodeAppropiateSecond() {
+		
+		Calendar calendar = Calendar.getInstance(); 
+		calendar.add(Calendar.SECOND, 99);
+		service.setAuthCodeExpirationSeconds(100);
+		ArgumentCaptor<AuthorizationCodeEntity> argumentCaptor = ArgumentCaptor.forClass(AuthorizationCodeEntity.class);
+		Mockito.verify(repository).save(argumentCaptor.capture());
+		assertThat(argumentCaptor.capture(), is(equalTo(calendar.getTime())));
```
jricher: As stated in previous comments, this should be doing a softer comparison than what it's doing right now. Instead of using `equalTo` make sure that it's within a 1-2 second window of the timestamp.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultOAuth2AuthorizationCodeService
```diff
@@ -0,0 +1,135 @@
...
+
+		}
```
jricher: Fix formatting throughout the test

PR #1367, opened 2018-02-23
Created few test cases for TestDefaultSystemScopeService

aishgosavi:


openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+		assertThat(service.save(defaultScope1), is(nullValue()));
+
+
+		
+	}
+	
+	@Test
+	public void getReserved() {
+		//Set<SystemScope> restricted = Sets.newHashSet(defaultScope1, defaultScope2, restrictedScope1);
+		Set<SystemScope> reserved = service.getReserved();
```
jricher: This is a pointless test because it's comparing a value to itself without making any changes in between. Better to copy or replicate the reserved scopes here so that if the reserved set changes in the future the unit tests will alert someone to that. Alternatively, you can pull the "reserved" set directly from the SystemScopeService interface instead of the instance object here, and then compare that to the instance below. Either way, calling the same function twice and getting the same result is a non-test.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+	}
+	
+	@Test
+	public void getremoveReservedScopes() {
+		Set<SystemScope> removeReservedScopes = Sets.newHashSet(defaultDynScope2, defaultDynScope1, defaultScope2,restrictedScope1,defaultScope1,dynScope1);
+		assertThat(service.removeReservedScopes(allScopes), equalTo(removeReservedScopes));
+	}
+	
+	@Test 
+	public void getById() {
```
jricher: Name these tests something more descriptive, like "getById_notFound".

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -175,7 +175,8 @@ public void shouldAssembleExpectedResultForAccessTokenWithoutExpiry() {
 				.build();
 		assertThat(result, is(equalTo(expected)));
 	}
-
+
```
jricher: The changes in this file are just whitespace, are you missing code?

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
 
 		// extra scope (fail)
 		assertThat(service.scopesMatch(expected, actualBad), is(false));
+		
 	}
+	
+	@Test
+	public void getremoveRestrictedAndReservedScopes() {
+						
+		Set<SystemScope> restrictedAndReservedScopes = Sets.newHashSet(defaultDynScope1, defaultDynScope2, dynScope1);
```
jricher: bad variable name: it looks here like you're making a set of just the restricted and removed scopes, where you're really making a list of unrestricted scopes

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
 
 		// extra scope (fail)
 		assertThat(service.scopesMatch(expected, actualBad), is(false));
+		
 	}
+	
+	@Test
+	public void getremoveRestrictedAndReservedScopes() {
```
jricher: Bad method name: there's no "get" involved here.

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+						
+		Set<SystemScope> restrictedAndReservedScopes = Sets.newHashSet(defaultDynScope1, defaultDynScope2, dynScope1);
+		assertThat(service.removeRestrictedAndReservedScopes(allScopes),equalTo(restrictedAndReservedScopes));
+	}
+	
+	@Test
+	public void getremoveReservedScopes() {
```
jricher: Same two comments as above

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+		
+		
+	}
+	
+	@Test
+	public void save_scope() {
+		
+		// check null condition 
+
+		assertThat(service.save(defaultScope1), is(nullValue()));
```
jricher: Better to have these as two separate tests

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+		
+		// check null condition
+		
+		assertThat(service.getByValue("defaultDynScope1String"), is(nullValue()));		
+		
+		
+	}
+	
+	@Test
+	public void save_scope() {
```
jricher: the _scope portion is unnecessary in this name, you're testing the "save" function. Better to have two: "save_null" and "save_value"

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultSystemScopeService.java
```diff
@@ -184,6 +184,60 @@ public void scopesMatch() {
...
+		assertThat(service.save(defaultScope1), is(nullValue()));
+		assertThat(service.save(mySystemScope), equalTo(mySystemScope));
+
+
+		
+	}
+	
+	@Test
+	public void getReserveScopes() {
+		Set<SystemScope> reserved = SystemScopeService.reservedScopes;
```
jricher: No need to pull this out as a separate variable, just reference it in the test below

aishgosavi:
I believe the build is passing now and the code is compiling. I have also addressed on the above comment and committed the changes to existing pull request.
Kindly guide me if this is correct way to do it.

aishgosavi:
Hi Justin,
I request you could you please have a look at my last commit and guide me over it.
I have passed userInfo as null paramter and i am using parameters which are  already used in the existing test cases.
I have taken references from other test cases which are similar to @test(expected= error.class) format.
Can you Kindly go through the test case and guide if this is the correct way and also on how to deal with parameter in this case.
Thank you

aishgosavi:
made changes in TestDefaultSystemScope, the build is failing because of the previous TestDefaultIntrospectionResultAssembler test case which is failing.

aishgosavi:
Hi Justin, I have made changes to the previous test cases and committed it. I have also commented in the code where I don't expect it to reach. Can you please go through it. Thank you

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -302,7 +306,75 @@ public void shouldAssembleExpectedResultForRefreshTokenWithoutUserAuthentication
...
+						.put("scope", "bar foo")
+						.put("active", Boolean.TRUE)
+						.put("user_id", "name")
+						.put("client_id", "clientId")
+						.put("token_type", "Bearer")
+						.build();
+				
+				
+				doThrow(ParseException.class).when(accessToken).getExpiration();
+				accessToken.getExpiration();
```
jricher: This is a pointless test. You're telling the mock to throw an exception when you make a call and then calling that function explicitly. 

What are you trying to accomplish here?

openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultIntrospectionResultAssembler.java
```diff
@@ -302,7 +306,75 @@ public void shouldAssembleExpectedResultForRefreshTokenWithoutUserAuthentication
...
+		Map<String, Object> expected = new ImmutableMap.Builder<String, Object>()
+				.put("sub", "sub")
+				.put("exp", 123L)
+				.put("scope", "bar foo")
+				.put("active", Boolean.TRUE)
+				.put("user_id", "name")
+				.put("client_id", "clientId")
+				.build();
+		
+		doThrow(ParseException.class).when(refreshToken).getExpiration();
```
jricher: Same comment as above. You're not supposed to be testing the mocked objects.

PR #1368, opened 2018-02-23
 Test cases for TestDefaultDeviceCodeService

Sangeeta19:


openid-connect-server/src/test/java/org/mitre/oauth2/service/impl/TestDefaultDeviceCodeService.java
```diff
@@ -161,19 +160,23 @@ public void testFindDeviceCodeFound() {
...
-/*
+
+	@SuppressWarnings("unchecked")
 	@Test
-	public void testClearExpiredDeviceCodes() {
+	public void testClearExpiredDeviceCodes() {		
+		expiredDeviceCodes1 = Sets.newHashSet(deviceCode1, deviceCode2);
+		expiredDeviceCodes2 = Sets.newHashSet();
+
+		Mockito.when(repository.getExpiredCodes()).thenReturn(expiredDeviceCodes1, expiredDeviceCodes1,expiredDeviceCodes2);
```
jricher: You're getting the test to pass here through some side effects. If the underlying paging system were to change, it might suddenly fail because of these side effects.

Sangeeta19: Justin,
Instead of invoking Mockito.veriyf() with hard code values like 2 and 4, invoking with size of set expiredDeviceCodes1 will solve the issue. 

Is my understanding of the your comment is correct?

Sangeeta19: will the below code snippet works in case if it is all about size:

Set<DeviceCode> expiredDeviceCodes1 = Sets.newHashSet(deviceCode1,deviceCode2); 
Set<DeviceCode> expiredDeviceCodes2 = Sets.newHashSet();
Mockito.when(repository.getExpiredCodes())
               .thenReturn(expiredDeviceCodes1,expiredDeviceCodes2);
service.clearExpiredDeviceCodes();

Mockito.verify(repository,times(expiredDeviceCodes1.size()))
                                       .remove(Matchers.any(DeviceCode.class));				
Mockito.verify(repository,times(expiredDeviceCodes1.size())).getExpiredCodes();

jricher: It's not about size, it's about how you're making your repository respond. The size calls are an improvement, but that's not the main issue. You're having your mocked repository return a full list twice and an empty list regardless of whether or not the expirations have been performed on it in the mean time. In other words, it's passing because the service thinks it's doing the right thing but it would still pass if something else a little weird happened.

Sangeeta19: By adding Sysout to method fetchPage() and doOperation(DeviceCode item) in implementation class, realized that for each call of fetchPage() : doOperation(DeviceCode item) calls are dependent on size of expiredCodesList, so all the items are removed. 

Then after fetchPage() is called again , now empty set will be returned (from mocked repository) as it got emptied out in the last call. 

This forced me to mock repository with list with elements for first call and zero elements in second call.

jricher: But that's only true if things are emptied out as expected with each page.

Sangeeta19:
Justin, I tried and I could not come up with the better way of mocking the repository to test the clearExpiredDeviceCodes(). Could you please help me.

Lionz01 (approved):

PR #1370, opened 2018-03-01
fix for CORS

dnsserver:
It is best not to continue on the filter chain in case of OPTIONS.


jricher:
Why is that?

dnsserver:
This is needed only for pure web browser apps ( no backend) on endpoints
that need authorization headers.
The OPTIONS request is strictly about establishing what
capabilities/permissions are valid for CORS interaction - there is
not/should not be any exchange of sensitive user data.
Therefore, if you apply filter chain even for OPTIONS, then users get 401
status response (even though the response headers are correct). This status
causes javascript to issue an error and not continue.

jricher:
Did you mean to add all these additional commits to this pull request? This is way more than adding a change to CORS now, and I'm inclined to close the pull without merging because it's doing a lot of other stuff.

dnsserver:
Nope, my bad. Should have sent the pr from it's own branch.

PR #1378, opened 2018-04-06
Fetch tokens by user name

ketola:
- In our environment, we don't delete users from db, we just mark them deleted and when this happens we need to delete the access and refresh tokens from the user
- There already was a method in DefaultOAuth2ProviderTokenService for getting the tokens for a user, but that reads all the tokens from database and loops through them to find the correct tokens, in our case the result from the query would be too big
- I created jpa queries to fetch the tokens by user name / sub and changed the implementation of DefaultOAuth2ProviderTokenService to use these new methods
- I didn't find a way to write tests for the repository from the existing unit tests but I did test it manually
- I also did some clean up in the classes I modified


jricher:
Overall this looks ok. In the past we've tried to avoid having too many passthrough methods to the repository layer but this one might make sense. And thank you for providing tests!

ketola:
Yes, I was initially not going to write tests for the repository as I saw that there were no similar tests made before. But I had to write some more tests for the codecov checks to pass. But anyway I think it's a good idea to have tests for the repository as there might be changes to the tables and queries that could cause bugs that would be noticed only by manual testing without the integration tests.

Issue #1384, opened 2018-04-19
"Unable to load locale" log should not be ERROR level


elennick:
If you make a request to the `/token` endpoint with an `Accept-Language` header that isn't supported by the server, the resulting log is ERROR level. This has resulted in our logs being spammed up with errors that should really be some lower level, probably just INFO.

So more specifically, if I make a token request and specify`Accept-Language: es` for Spanish or `Accept-Language: vi` for Vietnamese, then the following shows up in logs as part of the request at the ERROR level because there are no locale directories/files for those languages:

```j.i.FileNotFoundException: class path resource [resources/js/locale/es/messages.json] cannot be opened because it does not exist```

The code that generates this log statement is here: https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/e2d94f422a09a463854ff0919d560641cd06ef23/openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java#L186

I could create a PR to log as INFO for `FileNotFoundException` specifically and leave it as ERROR for all other checked exceptions like it currently is. Would that solution be reasonable?

jricher:
That sounds reasonable. Another thought is that the server should probably remember whether or not it's tried to load a file before. We cache positive results already but not negative ones. This would also cut down on the log messages you'd be getting over time.

PR #1385, opened 2018-04-21
"Unable to load locale" log should not be ERROR level 

elennick:
This is to resolve issue #1384 

If a locale message file is not found, it should now populate the `languageMap` with a null value and not try to load it every single time it gets requested. I also changed the resulting log message to INFO level instead of ERROR.

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
+++ b/openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
@@ -16,18 +16,6 @@
 
 package org.mitre.openid.connect.config;
 
-import com.google.common.base.Splitter;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonIOException;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
-import com.google.gson.JsonSyntaxException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.support.AbstractMessageSource;
-import org.springframework.core.io.Resource;
-
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -39,6 +27,19 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.support.AbstractMessageSource;
+import org.springframework.core.io.Resource;
+
+import com.google.common.base.Splitter;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonIOException;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.JsonSyntaxException;
 
 /**
  * @author jricher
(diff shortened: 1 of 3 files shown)


openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
```diff
@@ -147,16 +150,14 @@ private String getValue(String code, JsonObject lang) {
...
 		return value;
-
 	}
 
 	/**
 	 * @param locale
 	 * @return
 	 */
-	private List<JsonObject> getLanguageMap(Locale locale) {
+	List<JsonObject> getLanguageMap(Locale locale) {
```
jricher: Why did you make this package-private?

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
```diff
@@ -54,9 +54,13 @@
 
 	private Map<Locale, List<JsonObject>> languageMaps = new HashMap<>();
 
-	@Autowired
 	private ConfigurationPropertiesBean config;
 
+	@Autowired
```
jricher: We use `@Autowired` on fields instead of constructors throughout the rest of the project and I'd like to keep this somewhat consistent unless there's good reason to change just this one and not the others.

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
elennick: so that the unit test could access that method, see here: https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/pull/1385/files#diff-851c9e6cde51494f569f425d1cc0211bR37

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
elennick: It's a lot easier to insert test/mock objects into an object if you use constructor injection. The unit test I wrote requires no Spring context because it just constructs the config object as a POJO and passes it in during object construction. If you stick with field injection then you are tied to relying on the framework to construct objects correctly.

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
jricher: Please don't open up the class's signature just for unit tests. I understand the importance and desire for testing this method in isolation, but if that's the case you should either make it fully public and documented or use reflection to open it specifically for testing.

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
jricher: Other unit tests in the project use mock injection (with `@InjectMocks`) and it works fine.

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
elennick: fair enough... that method already has indirect coverage via testing of the public `resolveCode()` method so i removed the tests that directly test `getLanguageMap()`

openid-connect-server/src/main/java/org/mitre/openid/connect/config/JsonMessageSource.java
elennick: I just updated to keep field injection and inject using @InjectMocks. Let me know if that works.

PR #1386, opened 2018-04-23
Add support for verification_uri_complete

ondrejvelisek:
As draft suggests
https://tools.ietf.org/html/draft-ietf-oauth-device-flow-09#section-3.2

### DIFF
--- a/openid-connect-server/src/main/java/org/mitre/oauth2/web/DeviceEndpoint.java
+++ b/openid-connect-server/src/main/java/org/mitre/oauth2/web/DeviceEndpoint.java
@@ -16,6 +16,8 @@
 
 package org.mitre.oauth2.web;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
@@ -26,6 +28,7 @@ import java.util.UUID;
 
 import javax.servlet.http.HttpSession;
 
+import org.apache.http.client.utils.URIBuilder;
 import org.mitre.oauth2.exception.DeviceCodeCreationException;
 import org.mitre.oauth2.model.ClientDetailsEntity;
 import org.mitre.oauth2.model.DeviceCode;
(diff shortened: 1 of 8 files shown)


openid-connect-server/src/main/java/org/mitre/oauth2/web/DeviceEndpoint.java
```diff
@@ -139,6 +139,7 @@ public String requestDeviceCode(@RequestParam("client_id") String clientId, @Req
 			response.put("device_code", dc.getDeviceCode());
 			response.put("user_code", dc.getUserCode());
 			response.put("verification_uri", config.getIssuer() + USER_URL);
+			response.put("verification_uri_complete", config.getIssuer() + USER_URL + "?user_code=" + dc.getUserCode());
```
jricher: It makes me nervous not using a URL Builder for this. Yes, I know that the user code we're generating is already URL safe but I'd rather have it be explicitly handled in case someone replaces the user code generator in a custom service.

openid-connect-server/src/main/java/org/mitre/oauth2/web/DeviceEndpoint.java
```diff
@@ -160,12 +161,20 @@ public String requestDeviceCode(@RequestParam("client_id") String clientId, @Req
 
 	@PreAuthorize("hasRole('ROLE_USER')")
 	@RequestMapping(value = "/" + USER_URL, method = RequestMethod.GET)
-	public String requestUserCode(ModelMap model) {
+	public String requestUserCode(@RequestParam(value = "user_code", required = false) String userCode, ModelMap model, HttpSession session) {
```
jricher: Would it make sense to have a flag somewhere enabling this behavior? Perhaps per-client? There's no standard for that flag so maybe just adding it directly like this can work for now.

ondrejvelisek:
I've added requested changes (Not really sure about database stuffs, please check them)

jricher:
Thank you for this! We decided to back out the per-client switch right now because that would require further updates to the database and the UI. However, to prevent this feature from being on by default (as it does have some security and phishing concerns), we added a flag to the `ConfigurationPropertiesBean` that's set up in `server-config.xml` to turn it on and off across the whole server. It's off by default but should turn on just fine with setting that one new property.

PR #1405, opened 2018-06-11
Corrected Oracle create script

angelok1:
The oracle create script doesn't properly cleanup after itself if re-run.


jricher:
I'm not an oracle expert and have no means of testing this, so as with other Oracle DB stuff I'm going to have to trust that this is correct.

PR #1406, opened 2018-06-11
Fixed broken scripts from schema change on system_scope

angelok1:
When system_scope changed, the scope scripts for all databases needed to change as well.


jricher:
Thanks, we missed that on the 1.3 release somehow and this is the first complaint I've seen!

jricher:
Though FWIW, these scripts are really only meant to be used by temporary databases. Production systems should not be running these on startup.

angelok1:
Agreed. It's just nice to have if someone wants to bootstrap for development or have a reference implementation

PR #1409, opened 2018-06-20
Fix slow query for unused AuthenticationHolder entities (Postgres)

jansinger:
The named query for unused AuthenticationHolder entities uses "not in" selects resulting in slow query times (>30 min) on Postgres with a large amount of access token generations. Using "not exists" queries instead fixes this behavior.
Additionally the device code dependency is now checked as well to be sure that the AuthenticationHolder is not longer referenced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
 Unreleased:
+- Authorization codes are now longer
+- Client/RS can parse the "sub" and "user_id" claims in introspection response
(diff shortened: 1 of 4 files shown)


shrexster42:
Hi @jansinger , We are also seeing >15 min running time for the unused authentication holder named queries in our production systems. I will try out the solution you have mentioned. Thank you! 

+1 for merging this change to master

PR #1425, opened 2018-08-28
OpenID Connect Session Management

jansinger:
Implementation of OpenID Connect Session Management as defined in http://openid.net/specs/openid-connect-session-1_0.html.
Session State Management is enabled by default but can be disabled in the server configuration.
Uses the existing end session endpoint without modifications.
In addition to the standard the OP iFrame code re-validates the session state by calling a validation endpoint at the server after a configurable time (60s by default) to ensure that the session is refreshed and invalidated, e.g. by administrative logout or idle timeout.


shrexster42:
Hi @jansinger/ @jricher - any plans on merging this session management feature into the project?

PR #1429, opened 2018-10-06
2.0.0

kabiliravi:
In version 2.0.0 of OpenID Connect server, I added a feature to split the data for different proxy host endpoint pointing to the same OpenID Connect server which helps each data owner has access to see and update their own data. 
This feature enables OpenID server to be clustered and running in multiple instances, in a cloud environment under a Load Balancer.

Technically, that means if we have two host domain name one https://nasim.kabiliravi.com and the other one https://mitreid.org pointing to the same running server, their data completely separated.

I created a table called host_info table to map the host domain name with a UUID and I added a host_uuid field in order to all main tables (not nested tables). I also changed the type of primary key of all tables from auto incremental big integer to a randomly generated unified universal identifier version 4 (UUID v4) to have non trackable id for two consumers.

I am going to add more architectural and design diagrams to share my thoughts.


jricher:
Hi, we'll take a look at the code changes -- but first thing's first, you can't just declare this the "2.0.0-RELEASE" in the pull request. There's a release system that manages those version numbers.

jricher:
Have the data import/export classes been tested against data from older versions? The way it stands currently it doesn't look like they would successfully complete an import.

kabiliravi:
> Hi, we'll take a look at the code changes -- but first thing's first, you can't just declare this the "2.0.0-RELEASE" in the pull request. There's a release system that manages those version numbers.

I had issue with SNAPSHOT and making a executable standalone jar, I will definitely revert it back to SNAPSHOT. Also currently I am pointing to a nexus repo server that I am running in my host, I will revert back the distributionManagement and repository URLs to what they were

kabiliravi:
> Have the data import/export classes been tested against data from older versions? The way it stands currently it doesn't look like they would successfully complete an import.

The tests are passing, but I definitely will run the older instances locally, and I will test them manually by exporting them from older versions and importing them into the 2.0.0 server, to make sure everything works fine.

Please let me know if there is a better way of doing those tests or any other concerns.

jricher:
I guess something I'm not understanding from your base request is that if you've got multiple instances behind a load balancer, why go through all this trouble to separate the domains? Why not have the load balancer just dispatch to different clusters depending on the incoming domain, and have each of those clusters have their own internal data store?

kabiliravi:
> I guess something I'm not understanding from your base request is that if you've got multiple instances behind a load balancer, why go through all this trouble to separate the domains? Why not have the load balancer just dispatch to different clusters depending on the incoming domain, and have each of those clusters have their own internal data store?

Definitely that is one architectural solution specially for enterprise consumers which they can pay for the hardware resources and they want to run OpenID-Connect-Java-Spring-Server in their own infrastructure. However, for the consumers that they do not want to pay for a standalone server hardware resources and also they do not want the other consumers see their data, having their data filtered by a subdomain of their main domain such as http://oic.mycompany.com which proxies its requests/responses to a shared OpenID-Connect-Java-Spring-Server instance can be a desirable lean business product I believe. 
I am building a start-up based on the OpenID-Connect-Java-Spring-Server that you initiated. My plan is to build my own front-end using **react** or **angular** and use OpenID-Connect-Java-Spring-Server as the backend for Open ID and OAuth 2.
I also have some questions in regards to the OpenID-Connect-Java-Spring-Server license and copyright. I saw it is under Apache License but I wanted to be clear more about it.
I would be grateful if you send me an email to my personal email address: kabiliravi@gmail.com, so I can talk to you about that in details.

kabiliravi:
@jricher I fixed import and export data can you help me for the approval this pull request, do I need to anything else. I ran the server on http://kopenid.systeminventors.com/, it is a little bit slow because of the server spec.

kabiliravi:
@jricher Can you help me to have this pull approved? What should I do more?

srmoore:
So I've taken a look at the changes (which a majority of them seem to move from using the JPA autogenerated IDs to specifying a UUID...), and given the statement below I still have questions

> However, for the consumers that they do not want to pay for a standalone server hardware resources and also they do not want the other consumers see their data, having their data filtered by a subdomain of their main domain such as http://oic.mycompany.com which proxies its requests/responses to a shared OpenID-Connect-Java-Spring-Server instance can be a desirable lean business product I believe.

If you want to separate the data, then you shouldn't have them running through the same OIDC server. If you were setting up to be a OIDC service provider, wouldn't it be better to have a load balancer direct requests to separated pools of OIDC servers (most likely separate container instances) with their own datastores? Why complicated the OIDC server to do this filtering and looks ups in the database when you could just have two instances (even running on the same machine but different ports) that handle requests from domain A vs domain B?

I'm having trouble seeing the 'why' here.

I actually have a setup running on my machine here to support a similar setup because I'm running multiple projects, and each project has it's own OIDC instance (each a docker compose setup with it's own datastore) utilizing nginx to route requests between them based on request URI. In any case, what you are describing seems to be a special case, and really doesn't seem to be appropriate for the default versions of what is ultimately a reference implementation of the OIDC spec.

kabiliravi:
@srmoore The best reason I can answer 'why' or defend the architecture I propose
is going back to the old discussion of Monolithic vs. Microservices Architecture. 

Please take a look at this: https://github.com/kabiliravi/OpenID-Connect-Java-Spring-Server/blob/master/docs/cloud-arch-diagram.png

The current OpenID-Connect-Java-Spring-Server is following Monolithic Architecture. 
In the design I'm proposing I'm trying to make OpenID-Connect-Java-Spring-Server as thin as possible and make it a microservice which needs less hardware resources.

The ideal target is to take all the data access layer (including Default service classes, Default repository classes and JPA entities) out of the core project and let the core project (openid-connect-server) communicates with REST APIs for its data access layer.  
The beauty of this architecture is that, we can make those REST APIs configurable in runtime and based on the domain name we can dispatch the call to its corresponding REST API. (I presented that in the cloud-arch-diagram.png file)

The other layer is the presentation layer that exists in the core project (openid-connect-server). I mean @Controller classes located in this package specially (https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/tree/master/openid-connect-server/src/main/java/org/mitre/openid/connect/web)

srmoore:
I'm still not convinced that this is appropriate for the Default implementation classes. I feel the proper way to do this would be to create specialized versions of the Services and Repositories that can be optionally wired in via spring's autowire capability. I.E. a `RESTEnabledUserInfoService` that implements the `UserInfoService` interface, and prehaps bypasses the JPA `UserInfoRepository` class (instead caching the results in memory or in a Redis instance). Then you'd configure the spring webapp to use that instance rather than the Default. For instance, we've seen other instances of the project backed up to an LDAP server in this manner for deployment inside an enterprise.

The intent of the Default versions of the classes is to enable one to spin up the reference implementation without wiring into a bunch of specialized services.

kabiliravi:
@srmoore In summary, 
**Cons**:
1. The current architecture as a monolithic architecture consumes more hardware resources
2. Having the database transaction management inside this Monolithic server will cause database locks on heavy loads
3. Having the frontend inside this server increase the unnecessary workload of this server which can be separated as a complete different process
4. Having auto-incremental integer value for the tables primary keys which are being generated on database side is good for a small applications, but not when we are going to tie this server to one or more external REST mirecroservices. 
**Pros**:
Easier to develop and maintain

My final goal which I am currently work on it, is to make OpenID-Connect-Java-Spring-Server a spring boot microservice application without any front-end for data-entry and the only front-end it has is related to the OAuth 2 and OpenID such as login page and confirm access page and implement the front-end part in separate project which runs as a separate process. 
I also plan to move the Data Access Layer (Default Repositories) to a separate projects and based domain-driven design principles package them intro multiple REST domain microservices. 
Finally, to implement another plugin which contains REST proxy classes to fetch the data from the REST domain microservices or from the same contract REST APIs.

I find the implementation of this project very clean and the best source code that I can use, 
if the restructure and the re-factor I proposed is not aligned with the purpose of this project, I am not sure about is how to use this source codes if I am going partially use it or use them with a different package name without breaking the copyright and license. I know I have to create a separate project with a different name and have mitreid-connect Apache License inside my project. If I keep the license, on top the classes and codes I use, is it enough?

jricher:
I agree that this proposed set of changes is not within the spirit of the project, nor in a direction that I think most people would want to go with it. I still don't believe that your approach is a good way to tackle the problems you're trying to solve, as has been discussed at length above. 

Since the server is already entirely REST API driven, and the data layer is already separate from the rest of the system, what you're doing shouldn't be overwhelmingly difficult. 

You are free to include the source in your own project as per the terms of the Apache license. I'm not a lawyer, but as I understand it, it means that among other things it means you need to keep copyright and license notices in places where it already exists, and you cannot call your forked project "MITREid Connect" or a direct derivative of it (like MITREid Connect 2).

Issue #1468, opened 2019-03-13
CVE-2019-3778


jameskhedley:
Hi, given that this [vulnerability](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-3778) is quite severe and that the jar in question `spring-security-oauth2-2.1.0.RELEASE.jar` is a direct dependency in the [pom](https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server/blob/master/pom.xml), could the version please be upgraded to a fixed one, e.g. [2.1.4](https://mvnrepository.com/artifact/org.springframework.security.oauth/spring-security-oauth2/2.1.4.RELEASE) please?

Happy to fork and PR if necessary.

jricher:
Is this server actually vulnerable, though? We have a custom redirect resolver. Updating the dependency is simple enough (and a pull request would be appreciated for that) but I don't think we're susceptible to this particular bug.

jameskhedley:
Honestly I'm not sure. In our case our product was pulling it in which was causing our CI security scanner to complain. 

Anyway I made PR 1469, but the test are failing with 2.1.4.... Any ideas why?

jansinger:
See my comment on the PR. The "non strict" mode of the BlacklistAwareRedirectResolver causes this fail, because the DefaultRedirectResolver now also requires a strict match of the path, whereas before it excepts additional path elements.

This will potentially break all installations using the non strict or heart mode, as they need to register the exact redirect urls for all clients.

jameskhedley:
Oh, so it's a direct result of fixing the vuln then?

We can override the dependency locally so it's not really hurting us.

jansinger:
Yes, it is a direct result of the fix. Just override it locally and do not use the "non strict" mode, we are doing the same without any problem.

jameskhedley:
Fine by me, just FYI then that this project will be getting flagged.

blm126:
It looks like all of the pull requests to fix this got stalled because of failed unit tests related to non-strict mode of the BlacklistAwareRedirectResolver.  I'm willing to send a pull request, but it seems like there is an actual breaking change needed here.  As far as I understand, non-strict mode defaults to DefaultRedirectResolver which has changed behavior because of CVE-2019-3778. 

 Should non-strict mode go away or should it be re-implemented by BlacklistAwareRedirectResolver to preserve a feature?

PR #1469, opened 2019-03-14
bumped oauth2 dependency to 2.1.4 because of CVE-2019-3778

jameskhedley:
issue #1468

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -393,7 +393,7 @@
 			<dependency>
 				<groupId>org.springframework.security.oauth</groupId>
 				<artifactId>spring-security-oauth2</artifactId>
-				<version>2.1.0.RELEASE</version>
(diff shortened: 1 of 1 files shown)


jansinger:
The fixed DefaultRedirectResolver of Spring Security now does a strict path match as required, so that the non strict mode of the BlacklistAwareRedirectResolver does not work as expected anymore.
My suggestion would be to remove the non strict mode to comply with the security requirements, but this will break existing installations using this feature, especially those using the heart mode.

jricher:
FWIW, Heart mode requires strict matching, and strict mode has been the default since 1.3.2 anyway with or without heart mode.

jricher:
This pull request needs to fix the associated unit tests if it's to be included.

PR #1496, opened 2019-07-10
Update Spring to 5.1.8.RELEASE and spring-security-oauth2 to 2.3.6.RELEASE

angelusGJ:


openid-connect-client/src/main/java/org/mitre/openid/connect/client/service/impl/StaticClientConfigurationService.java
```diff
@@ -66,7 +65,7 @@ public RegisteredClient getClientConfiguration(ServerConfiguration issuer) {
 		return clients.get(issuer.getIssuer());
 	}
 
-	@PostConstruct
+	@Override
```
jricher: Why was this change made? We moved to using PostConstruct many years ago, so I'm not sure why we'd move back.

PR #1498, opened 2019-07-19
Fix error-page for POST,PUT, etc.

thomasmey:
the view controller error in authz-config.xml forwards to the resp. JSP,
but the error page only supports GET method.

Install a real error controller that at least logs all error messages
and then forwards to the view controller, so not handled exceptions for
e.g. POST request can actually be found.

### DIFF
--- a/openid-connect-server-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/openid-connect-server-webapp/src/main/webapp/WEB-INF/web.xml
@@ -71,9 +71,9 @@
 			<trim-directive-whitespaces>true</trim-directive-whitespaces>
 		</jsp-property-group>
 	</jsp-config>
-	
+
 	<error-page>
-		<location>/error</location>
+		<location>/errorController</location>
 	</error-page>
-	
+
 </web-app>
--- a/pom.xml
+++ b/pom.xml
@@ -404,8 +404,8 @@
 			</dependency>
 			<dependency>
 				<groupId>javax.servlet</groupId>
-				<artifactId>servlet-api</artifactId>
-				<version>2.5</version>
+				<artifactId>javax.servlet-api</artifactId>
+				<version>3.0.1</version>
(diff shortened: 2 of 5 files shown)


openid-connect-server-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -71,9 +71,9 @@
 			<trim-directive-whitespaces>true</trim-directive-whitespaces>
 		</jsp-property-group>
 	</jsp-config>
-	
+
 	<error-page>
-		<location>/error</location>
+		<location>/errorController</location>
```
jricher: Can we still map this to `/error`?

pom.xml
```diff
@@ -404,8 +404,8 @@
 			</dependency>
 			<dependency>
 				<groupId>javax.servlet</groupId>
-				<artifactId>servlet-api</artifactId>
-				<version>2.5</version>
+				<artifactId>javax.servlet-api</artifactId>
+				<version>3.0.1</version>
```
jricher: Are there any other changes that updating the servlet version would bring to the underlying project?

openid-connect-server-webapp/src/main/webapp/WEB-INF/web.xml
thomasmey: So I'm not sure this is possbile without breaking things. Some other jsp pages directly redirect/forwards to /error when changing the above "catch all" error location these wont get catch. Or do I misunderstand. Let me ask the otherway around:
- under what path should the ErrorController be reachable?
- under what path should the error.jsp be reachable?

another thing that should be done is to move the controller logic from error.jsp in the ErrorController.

pom.xml
thomasmey: this gives the static defines like RequestDispatcher.ERROR_MESSAGE as web.xml uses 3.0 syntax already.

PR #1527, opened 2020-02-19
Sanitize user names in model

### LINKED ISSUE #1521: Cross-Site Scripting header.tag
`header.tag` appears to be vulnerable to XSS here:

~~~
		// get the info of the current user, if available (null otherwise)
    	function getUserInfo() {
    		return ${userInfoJson};
    	}
~~~

**userInfoJson** is included in the page and is not encoded so malicious elements could be created.  If the string `</script>` appears in **userInfoJson**, the `<script>` element will be closed 
and a new malicious `<script>` can be created:
~~~
        // get the info of the current user, if available (null otherwise)
    	function getUserInfo() {
    		return {"sub":"12318767","name":"Test</script><script>alert(1)</script> Test","preferred_username":"Test","given_name":"Test</script><script>alert(1)</script>","family_name":"Test","email":"test@test.com","email_verified":true};
    	}
~~~
openid_xss2

And the malicious JavaScript is executed:

openid_xss

irbishop:
Something like this should fix #1521, but preserves sanity in the JSON:

fix

fix

### DIFF
--- a/openid-connect-common/src/main/java/org/mitre/openid/connect/model/DefaultUserInfo.java
+++ b/openid-connect-common/src/main/java/org/mitre/openid/connect/model/DefaultUserInfo.java
@@ -37,6 +37,7 @@ import javax.persistence.Table;
 
 import org.mitre.openid.connect.model.convert.JsonObjectStringConverter;
 
+import com.google.common.html.HtmlEscapers;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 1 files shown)


jricher:
Since it deals with escaping at the data object level, this approach would break the use of the value in other parts of the server. This also wouldn't help anyone with a custom user info class.

PR #1554, opened 2021-06-03
Fix css admin

dBucik:


dBucik:
Sorry, opened to wrong repo.

